PR #4, opened 2016-02-03
Implement first method transfer manager

kyleknap:
Has the ability to upload files to s3 by providing the filename.

Sorry that it is really long. I am hoping that with the way in which I designed the interface of the internals, it is going to take a significant less amount of effort and code to add the other methods/functionality. It just did not make sense to me break the internals up into separate PR's without any knowledge/reference of how I was going to use them. I would be more than happy to explain how everything fits together in person to help with the review.

cc @jamesls @mtdowling @rayluo @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -1,0 +1,244 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from concurrent import futures
+import functools
+import threading
+
+
+class TransferFuture(object):
+    def __init__(self, meta=None, context=None):
+        """The future associated to a submitted transfer request
+
+        :type meta: TransferMeta
+        :param meta: The metadata associated to the request. This object
+            is visible to the requester..
+
+        :type context: TransferContext
+        :param context: The context associated to the request. This object
+            is not visible to the requester.
+        """
+        self._meta = meta
+        if meta is None:
+            self._meta = TransferMeta()
+
+        self._context = context
+        if context is None:
+            self._context = TransferContext()
+
+    @property
+    def meta(self):
+        """The metadata associated tio the TransferFuture"""
+        return self._meta
+
+    def done(self):
+        """Determines if a TransferFuture has completed
+
+        :returns: True if completed. False, otherwise.
+        """
+        return self._context.done()
+
+    def result(self):
+        """Waits until TransferFuture is done and returns the result
+
+        If the TransferFuture succeeded, it will return the result. If the
+        TransferFuture failed, it will raise the exception associated to the
+        failure.
+        """
+        return self._context.result()
+
+    def cancel(self):
+        """Cancels the request associated with the TransferFuture"""
+        self._context.cancel()
+
+
+class TransferMeta(object):
+    """Holds metadata about the TransferFuture"""
+    def __init__(self, call_args=None):
+        self._call_args = call_args
+        self._size = None
+        self._user_context = {}
+
+    @property
+    def call_args(self):
+        """The call args used in the transfer request"""
+        return self._call_args
+
+    @property
+    def size(self):
+        """The size of the transfer request if known"""
+        return self._size
+
+    @property
+    def user_context(self):
+        """A dictionary that requesters can store data in"""
+        return self._user_context
+
+    def provide_transfer_size(self, size):
+        """A method to provide the size of a transfer request
+
+        By providing this value, the TransferManager will not try to
+        call HeadObject or use the use OS to determine the size of the
+        transfer.
+        """
+        self._size = size
+
+
+class TransferContext(object):
+    """A helper class for managing TransferFuture"""
+    def __init__(self):
+        self._status = 'queued'
+        self._result = None
+        self._exception = None
+        self._failure_cleanups = []
+        self._done_event = threading.Event()
+        self._lock = threading.Lock()
+
+    @property
+    def failure_cleanups(self):
+        """The list of callbacks to call when the TransferFuture fails"""
+        return self._failure_cleanups
+
+    @property
+    def status(self):
+        """The status of the TransferFuture
+
+        The currently supported states are:
+            * queued - Has yet to start
+            * running - Is inprogress
+            * cancelled - Was cancelled
+            * failed - An exception other than CancelledError was thrown
+            * success - No exceptions were thrown and is done.
+        """
+        with self._lock:
+            return self._status
+
+    def set_result(self, result):
+        """Set a result for the TransferFuture
+
+        Implies that the TransferFuture succeeded.
+        """
+        with self._lock:
+            self._result = result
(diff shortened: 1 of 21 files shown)


kyleknap:
Forgot that this botocore PR for stubbing:https://github.com/boto/botocore/pull/784  needs to be merged first before tests can pass...

kyleknap:
I did some initial testing between the transfer manager and the CLI ``cp`` command. Here it was I got for the speed of uploading a single 10GB file:
```
s3 cp
1m55.227s
1m33.227s
1m32.077s
1m33.663s

s3transfer
1m31.954s
1m32.433s
1m33.514s
1m38.684s
```
So the times looks similar which is good. The amount of memory being used and CPU being used were roughly the same using ``top``. I am going to next look at a directory with a bunch of files and see how that looks.

kyleknap:
Hmmm. Uploading many small files is pretty slow compared to the CLI. Uploading 10,000 files of 10kb size is taking about 1m10s with the ``cp`` commands and the ``s3transfer`` is currently like 5 to 7 times slower...

The good news is that a pure futures implementation is not the reason. This code gets me less than 1m10s:
```py
def upload_file(client, filename, bucket, key):
    with open(filename, 'rb') as f:
        client.put_object(Body=f, Bucket=bucket, Key=key)

def upload_many_files_with_client(client):
    bucket = 'mybucketfoo'
    file_list = glob.glob('many-files/*/*/*')
    with ThreadPoolExecutor(max_workers=10) as executor:
        for name in file_list:
            future = executor.submit(
                upload_file, client, filename=name, bucket=bucket, key=name)
```
So it is probably something to do with the implementation. Have yet to figure that out.

kyleknap:
It is the use of ``ReadFileChunk`` class. Wrapping the body in that class makes it much slower than just using the normal ``open()`` function with a context handler. When I remove all of the ``ReadFileChunk`` logic, the speed of s3transfer is about 1m10s, the same as the CLI.

jamesls:
Do you have a profile of what in the ReadFileChunk is slow?  Are there subscribers to the read callbacks that cause this slow down?

kyleknap:
Possibly. That is what I am trying to figure out next. In my perf script, I have no progress callbacks registered.

kyleknap:
Specifically, it looks like the slowness is caused by the registering and unregistering of the enabling of callbacks. Having to register and unregister an unique event to a single event emitter really bogs down the speed of the individual threads. When I remove the logic where I register and unregister the handlers I achieve a speed a bit faster than the CLI.

I am tempted to remove the dependency on ``ReadFileChunk`` as it has already been problematic in ``boto3`` such as with the sigv4 workaround and try and find another way to determine progress that revolves around socket or HTTP connection reading.

Another thing that I found that opening files in the main thread through ``ReadFileChunk`` causes IOErrors saying too many open files are present. Moving the file opening to each individual thread fixes this issue.

jamesls:
@kyleknap Is this related to the naive cache invalidation we have when unregistering handlers?  Right now in botocore, any modifications to the handler registrations unconditionally clears the cache.

kyleknap:
It maybe the reason. I have not looked too deep into it. But it would make sense if there were thousands of handlers being registered with no cache so you would have to search through them from scratch.

Honestly, I want to figure out a way to get progress without registering handlers. So this logic should hopefully be temporary as I backport boto3/CLI functionality to s3transfer.

JordonPhillips:
Looks good to me, pending the updates mentioned above.

s3transfer/futures.py
```diff
@@ -0,0 +1,246 @@
...
+
+        :returns: The future assocaited to the submitted task
+        """
+        with self._lock:
+            # First determine if there are already too many futures running
+            if self._at_maximum_running_futures():
+                # If there are too many futures running, wait till some
+                # complete and save the remaining running futures as the
+                # next set to wait for.
+                self._currently_running_futures = futures.wait(
```
jamesls: Do you know if this is creating a copy of the incomplete waiters each time?

I'm wondering about the case where the input rate is greater than the output rate which will mean we're essentially calling `.wait()` every time we try to submit.  If the max size is large enough, and we are creating a copy of incomplete futures every `wait()` call, this seems like this could create a lot of overhead.

Thoughts?

s3transfer/manager.py
```diff
@@ -0,0 +1,165 @@
...
+        :param key: The name of the key to upload to
+
+        :type extra_args: dict
+        :param extra_args: Extra arguments that may be passed to the
+            client operation
+
+        :type subscribers: a list of subscribers
+        :param subscribers: The list of subscribers to be invoked in the
+            order provided based on the event emit during the process of
+            the transfer request.
```
jamesls: I think it would be worth having a base class that a user can implement that would serve as documentation for the valid notifications you can subscribe to.

s3transfer/manager.py
```diff
@@ -0,0 +1,165 @@
...
+        call_args = CallArgs(
+            fileobj=fileobj, bucket=bucket, key=key, extra_args=extra_args,
+            subscribers=subscribers
+        )
+        upload_submitter = UploadTaskSubmitter(
+            client=self._client, osutil=self._osutil, config=self._config,
+            executor=self._executor)
+        return upload_submitter(call_args)
+
+    def _validate_subscribers(self, subscribers):
```
jamesls: Another benefit of creating a base class is that you can move all this validation to the metaclass so we don't have to validate the object at upload time.  A given class will either validate or not validate, and if a user wants to reuse the same subscriber instance each time we're unnecessarily validating the class repeatedly.

s3transfer/tasks.py
```diff
@@ -0,0 +1,245 @@
...
+        transfer_future = TransferFuture(**components)
+        on_queued_callbacks = get_callbacks(transfer_future, 'queued')
+        for on_queued_callback in on_queued_callbacks:
+            on_queued_callback()
+        return transfer_future, components
+
+    def __call__(self, call_args=None):
+        """Initializes and submits a transfer requests corresponding tasks
+
+        :param call_args: The call arguments associated to the request
```
jamesls: Can we add a type for this?  I had trouble working out what this is.

jamesls:
I've gone through everything except the tests, and so far I think this looks good.  It's pretty straightforward to follow, and seems like a big improvement over what we have currently.

I'll go through the tests shortly.

kyleknap:
I updated the PR with the following:
* Improved performance on uploading many files. It is now at the speed of the CLI.
* Made a ``BaseSubscriber`` class. It is a much better approach from a functionality and documentation perspective than before.

Should be good to look at again along with the tests.

kyleknap:
@jamesls so I just implemented to upload streams to s3: https://github.com/boto/s3transfer/pull/11. If you have not gotten too far in this PR (and I have a decent understanding of the internals), I would suggest that you take a look at that PR because I did a fair amount of refactoring in both the code and the tests. In the end, I feel that because of the refactoring I did, the implementation is better abstracted making it easier to read, understand, and test.

jamesls:
:shipit: Looks good.  I'll take a look at the streaming upload.

iamahuman:
> Another thing that I found that opening files in the main thread through `ReadFileChunk` causes IOErrors saying too many open files are present. Moving the file opening to each individual thread fixes this issue.

I suspect the commit (supposedly) fixing this issue, boto/s3transfer@2f3d12cba458d7f90e583c754e317e479c49c8c9, actually introduced regression described in #80.

It should be noted that the fixing commit introduced an additional behavior of closing the file argument. I suspect the thread from which the files were opened were not the issue by itself, but simply interfered the chances of the `EMFILE` issue cropping up due to concurrency of threads.

@kyleknap: Would it be possible that the main issue was, in fact, due to the (stress) test code not assuming the responsibility of closing the files itself, resulting in file descriptor leakage manifesting as `OSError: Too many open files (EMFILE)`?

PR #5, opened 2016-02-22
Implement download method for TransferManager

kyleknap:
cc @jamesls @JordonPhillips 

### Review Progress

- [x] ``s3transfer/compat.py``
- [x] ``s3transfer/download.py``
- [x] ``s3transfer/exceptions.py``
- [x] ``s3transfer/futures.py``
- [x] ``s3transfer/manager.py``
- [x] ``s3transfer/utils.py``
- [x] ``tests/__init__.py``
- [x] ``tests/functional/test_download.py``
- [x] ``tests/integration/__init__.py``
- [x] ``tests/integration/test_download.py``
- [x] ``tests/unit/test_download.py``
- [x] ``tests/unit/test_upload.py``
- [x] ``tests/unit/test_utils.py``

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -1,0 +1,294 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import functools
+import logging
+import os
+import socket
+import math
+
+from botocore.exceptions import IncompleteReadError
+from botocore.vendored.requests.packages.urllib3.exceptions import \
+    ReadTimeoutError
+
+from s3transfer.compat import SOCKET_ERROR
+from s3transfer.exceptions import RetriesExceededError
+from s3transfer.exceptions import QueueShutdownError
+from s3transfer.utils import random_file_extension
+from s3transfer.utils import get_callbacks
+from s3transfer.utils import StreamReaderProgress
+from s3transfer.utils import ShutdownQueue
+from s3transfer.tasks import Task
+from s3transfer.tasks import TaskSubmitter
+
+
+logger = logging.getLogger(__name__)
+
+SHUTDOWN_SENTINEL = object()
+S3_RETRYABLE_ERRORS = (
+    socket.timeout, SOCKET_ERROR, ReadTimeoutError, IncompleteReadError
+)
+
+
+class DownloadTaskSubmitter(TaskSubmitter):
+    def __init__(self, client, config, osutil, executor, io_executor):
+        super(DownloadTaskSubmitter, self).__init__(
+            client, config, osutil, executor)
+        self._io_executor = io_executor
+
+    def _submit(self, transfer_future, transfer_coordinator):
+        if transfer_future.meta.size is None:
+            # If a size was not provided figure out the size for the
+            # user.
+            response = self._client.head_object(
+                Bucket=transfer_future.meta.call_args.bucket,
+                Key=transfer_future.meta.call_args.key,
+                **transfer_future.meta.call_args.extra_args
+            )
+            transfer_future.meta.provide_transfer_size(
+                response['ContentLength'])
+
+        # Figure out a temporary filename for the file.
+        temp_filename = (
+            transfer_future.meta.call_args.fileobj + os.extsep +
+            random_file_extension()
+        )
+        # If it is greater than threshold do a ranged download, otherwise
+        # do a regular GetObject download.
+        if transfer_future.meta.size < self._config.multipart_threshold:
+            self._submit_download_request(
+                temp_filename, transfer_future, transfer_coordinator)
+        else:
+            self._submit_ranged_download_request(
+                temp_filename, transfer_future, transfer_coordinator)
+
+    def _submit_download_request(
+            self, temp_filename, transfer_future, transfer_coordinator):
+        call_args = transfer_future.meta.call_args
+
+        # Get the needed callbacks for the task
+        progress_callbacks = get_callbacks(transfer_future, 'progress')
+        done_callbacks = get_callbacks(transfer_future, 'done')
+
+        # Create the io queue that will be used for this download
+        io_queue = ShutdownQueue(self._config.max_io_queue_size)
+
+        # Start up the io task to write the downloads to disk as they
+        # come in.
+        self._submit_io_write_task(
+            temp_filename, call_args.fileobj, transfer_coordinator,
+            io_queue, done_callbacks)
+
+        # Submit the task to download the object.
+        self._executor.submit(
+            GetObjectTask(
+                transfer_coordinator=transfer_coordinator,
+                main_kwargs={
+                    'client': self._client,
+                    'bucket': call_args.bucket,
+                    'key': call_args.key,
+                    'extra_args': call_args.extra_args,
+                    'callbacks': progress_callbacks,
+                    'max_attempts': self._config.num_download_attempts,
+                    'io_queue': io_queue,
+                },
+                done_callbacks=[
+                    functools.partial(io_queue.put, SHUTDOWN_SENTINEL)
+                ]
+            )
+        )
+
+    def _submit_ranged_download_request(
+            self, temp_filename, transfer_future, transfer_coordinator):
+        call_args = transfer_future.meta.call_args
+
+        # Get the needed callbacks for the task
+        progress_callbacks = get_callbacks(transfer_future, 'progress')
+        done_callbacks = get_callbacks(transfer_future, 'done')
+
+        # Determine the number of parts
+        part_size = self._config.multipart_chunksize
+        num_parts = int(
+            math.ceil(transfer_future.meta.size / float(part_size)))
+
+        # Create the io queue that will be used for this download
+        io_queue = ShutdownQueue(self._config.max_io_queue_size)
+
+        # Start up the io task to write the downloads to disk as they
+        # come in.
+        self._submit_io_write_task(
+            temp_filename, transfer_future.meta.call_args.fileobj,
+            transfer_coordinator, io_queue, done_callbacks)
+
+        ranged_downloads = []
+        for i in range(num_parts):
+            # Calculate the range parameter
+            range_parameter = self._calculate_range_param(
+                part_size, i, num_parts)
+
+            # Inject the Range parameter to the parameters to be passed in
+            # as extra args
+            extra_args = {'Range': range_parameter}
+            extra_args.update(call_args.extra_args)
+
+            # Submit the ranged downloads
+            ranged_downloads.append(
+                self._executor.submit(
+                    GetObjectTask(
+                        transfer_coordinator=transfer_coordinator,
+                        main_kwargs={
+                            'client': self._client,
+                            'bucket': call_args.bucket,
+                            'key': call_args.key,
+                            'extra_args': extra_args,
+                            'callbacks': progress_callbacks,
+                            'max_attempts': self._config.num_download_attempts,
+                            'io_queue': io_queue,
+                            'start_index': i * part_size
+                        }
+                    )
+                )
+            )
+
+        # Submit a task to wait for all of the downloads to complete
+        # before telling the io task to stop writing.
+        self._executor.submit(
+            RangeDownloadJoinTask(
+                transfer_coordinator=transfer_coordinator,
+                pending_main_kwargs={
+                    'ranged_downloads': ranged_downloads
+                },
+                done_callbacks=[
+                    functools.partial(io_queue.put, SHUTDOWN_SENTINEL)
+                ]
+            )
+        )
+
+    def _submit_io_write_task(self, temp_filename, filename,
+                              transfer_coordinator, io_queue, done_callbacks):
+        self._io_executor.submit(
+            IOWriteTask(
+                transfer_coordinator=transfer_coordinator,
+                main_kwargs={
+                    'filename': temp_filename,
+                    'final_filename': filename,
+                    'osutil': self._osutil,
+                    'io_queue': io_queue
+                },
+                done_callbacks=done_callbacks,
+                is_final=True
+            )
+        )
+
+    def _calculate_range_param(self, part_size, part_index, num_parts):
+        # Used to calculate the Range parameter
+        start_range = part_index * part_size
+        if part_index == num_parts - 1:
+            end_range = ''
+        else:
+            end_range = start_range + part_size - 1
+        range_param = 'bytes=%s-%s' % (start_range, end_range)
+        return range_param
+
+
+class GetObjectTask(Task):
+    STREAM_CHUNK_SIZE = 8 * 1024
+
+    def _main(self, client, bucket, key, extra_args, callbacks,
+              max_attempts, io_queue, start_index=0):
+        """Downloads an object and places content into io queue
+
+        :param client: The client to use when calling GetObject
+        :param bucket: The bucket to download from
+        :param key: The key to download from
+        :param exta_args: Any extra arguements to include in GetObject request
+        :param callbacks: List of progress callbacks to invoke on download
+        :param max_attempts: The number of retries to do when downloading
+        :param io_queue: The queue that content of the object will be
+            placed in
+        :param start_index: The location in the file to start writing the
+            content of the key to.
+        """
+        last_exception = None
+        for i in range(max_attempts):
+            try:
+                response = client.get_object(
+                    Bucket=bucket, Key=key, **extra_args)
+                streaming_body = StreamReaderProgress(
+                    response['Body'], callbacks)
+
+                current_index = start_index
+                chunks = iter(
+                    lambda: streaming_body.read(self.STREAM_CHUNK_SIZE), b'')
+                for chunk in chunks:
+                    # If the io queue was shutdown just stop sending to
+                    # chunks to the queue and quit the download.
+                    try:
+                        io_queue.put((current_index, chunk))
+                    except QueueShutdownError:
+                        return
+                    current_index += len(chunk)
+                return
+            except S3_RETRYABLE_ERRORS as e:
+                logger.debug("Retrying exception caught (%s), "
+                             "retrying request, (attempt %s / %s)", e, i,
+                             max_attempts, exc_info=True)
+                last_exception = e
+                continue
+        raise RetriesExceededError(last_exception)
+
+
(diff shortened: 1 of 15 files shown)


JordonPhillips:
Nothing immediately pops out, it seems pretty well in line with the first PR.

kyleknap:
@jamesls @JordonPhillips I rebased off of develop should be good to look at.

s3transfer/download.py
```diff
@@ -0,0 +1,294 @@
...
+                for chunk in chunks:
+                    # If the io queue was shutdown just stop sending to
+                    # chunks to the queue and quit the download.
+                    try:
+                        io_queue.put((current_index, chunk))
+                    except QueueShutdownError:
+                        return
+                    current_index += len(chunk)
+                return
+            except S3_RETRYABLE_ERRORS as e:
```
jamesls: How are we handling callbacks on retry?  How would a callback handle something like displaying progress percentage?  It would have to know when retries happen right?

s3transfer/utils.py
```diff
@@ -241,3 +253,47 @@ def __iter__(self):
...
+        # queue.Queue is an old style class so we don't use super().
+        return queue.Queue._init(self, maxsize)
+
+    def trigger_shutdown(self):
+        with self._shutdown_lock:
+            self._shutdown = True
+            logger.debug("The IO queue is now shutdown.")
+
+    def put(self, item):
+        # Note: this is not sufficient, it's still possible to deadlock!
```
jamesls: How is deadlock possible?  What are the plans to address this?

tests/functional/test_download.py
```diff
@@ -0,0 +1,315 @@
...
+
+        # The HeadObject should have not happened and should have been able
+        # to successfully download the file.
+        self.stubber.assert_no_pending_responses()
+        with open(self.filename, 'rb') as f:
+            self.assertEqual(self.content, f.read())
+
+
+class TestNonRangedDownload(BaseDownloadTest):
+    __test__ = True
```
jamesls: What's `__test__` used for?

s3transfer/futures.py
```diff
@@ -128,8 +128,9 @@ def set_result(self, result):
         Implies that the TransferFuture succeeded.
         """
         with self._lock:
-            self._result = result
-            self._status = 'success'
+            if self._exception is None:
```
jamesls: What's this change for?

tests/integration/test_download.py
```diff
@@ -0,0 +1,69 @@
...
+            'foo.txt', filesize=20 * 1024 * 1024)
+        self.upload_file(filename, '20mb.txt')
+
+        download_path = os.path.join(self.files.rootdir, '20mb.txt')
+
+        future = transfer_manager.download(
+            self.bucket_name, '20mb.txt', download_path,
+            subscribers=[subscriber])
+        future.result()
+        self.assertEqual(subscriber.calculate_bytes_seen(), 20 * 1024 * 1024)
```
jamesls: Will this test fail if the download is retried?

s3transfer/futures.py
```diff
@@ -128,8 +128,9 @@ def set_result(self, result):
...
+                self._result = result
+                self._status = 'success'
```
jamesls: Tangentially related, but I noticed you can cancel a task after it's completed.  Is that intentional?  Seems like we should treat cancelled as a terminal state and not allow any transitions out of that state.

s3transfer/download.py
```diff
@@ -0,0 +1,294 @@
...
+        self._transfer_coordinator.add_failure_cleanup(
+            self._remove_file_cleanup, filename, osutil)
+        try:
+            self._loop_on_io_writes(filename, osutil, io_queue)
+        except Exception as e:
+            logger.debug("Caught exception in IO thread: %s",
+                         e, exc_info=True)
+            io_queue.trigger_shutdown()
+            raise
+        if self._transfer_coordinator.status != 'failed':
```
jamesls: This should also account for cancelled tasks, otherwise we're clobbering the original file.

Perhaps it's worth tracking terminal and/or failed states as class vars to make this more clear.

s3transfer/download.py
```diff
@@ -0,0 +1,294 @@
...
+                if task is SHUTDOWN_SENTINEL:
+                    logger.debug("Shutdown sentinel received in IO handler, "
+                                 "shutting down IO handler.")
+                    return
+                else:
+                    offset, data = task
+                    f.seek(offset)
+                    f.write(data)
+
+    def _remove_file_cleanup(self, filename, osutil):
```
jamesls: I noticed interesting behavior when `.result()` is not called.  We still block until the download finishes, but if you Ctrl-C the process, we don't have any hooks to run the cleanup.  So for example, you can leave temp files hanging around with something like this:

```
import boto3
import time

s3 = boto3.client('s3')
m = TransferManager(s3)
r = m.download('bucket', 'large', '/tmp/largefile')

print "Sleeping..."
time.sleep(3)
print "Hit ctrl-c now"
```

You'll see a `/tmp/largefile.*` file still hanging around.  What options (if any) do we have for handling this?

s3transfer/manager.py
```diff
@@ -42,14 +45,31 @@ def __init__(self,
         :param max_queue_size: The maximum amount of requests that
             can be queued at a time. A value of zero means that there
             is no maximum.
+
+        :param max_io_queue_size: The maximum amount of read parts that
+            can be queued to be written to disk per download. A value of zero
+            means that there is no maximum.
+
+        :param num_download_attempts: The number of download attempts that
+            will be tried upon errors with downloading an object in S3.
```
jamesls: Can we update the docs to say more retries can happen based on if they happen while writing the file out to disk vs. in the initial service response.  For each GetObject attempt in s3transfer, botocore could retry things like 5xx responses N times before handing the response over to s3transfer.

Not sure exactly how to word it, but something that lets them know there's more retries happening depending on where errors happen.

s3transfer/manager.py
```diff
@@ -127,6 +158,54 @@ def upload(self, fileobj, bucket, key, extra_args=None, subscribers=None):
...
+        :param key: The name of the key to download from
+
+        :type fileobj: str
+        :param fileobj: The name of a file to download to.
+
+        :type extra_args: dict
+        :param extra_args: Extra arguments that may be passed to the
+            client operation
+
+        :type subscribers: a list of subscribers
```
jamesls: Can we document the type of subscribers?

s3transfer/download.py
```diff
@@ -0,0 +1,294 @@
...
+        # Get the needed callbacks for the task
+        progress_callbacks = get_callbacks(transfer_future, 'progress')
+        done_callbacks = get_callbacks(transfer_future, 'done')
+
+        # Create the io queue that will be used for this download
+        io_queue = ShutdownQueue(self._config.max_io_queue_size)
+
+        # Start up the io task to write the downloads to disk as they
+        # come in.
+        self._submit_io_write_task(
```
jamesls: Does this mean that there's 1 IO queue and IO thread writer *per* download?

tests/functional/test_download.py
```diff
@@ -0,0 +1,315 @@
...
+        future.result()
+        # Make sure the file exists
+        self.assertTrue(os.path.exists(self.filename))
+        # Make sure the random temporary file does not exist
+        possible_matches = glob.glob('%s*' % self.filename + os.extsep)
+        self.assertEqual(possible_matches, [])
+
+    def test_download_cleanup_on_failure(self):
+        # Add the stubbed HeadObject response
+        self.stubber.add_response(**self.stubbed_responses[0])
```
jamesls: Rather than duplicating this comment and `stubbed_responses[0]` for each test, what about just having a `self.add_head_object_response()`.   Even better to provide the size in the head object.  It was hard tracking down responses[0] and then tracking down self.content to see how the size compared to the threshold.

tests/functional/test_download.py
```diff
@@ -0,0 +1,315 @@
...
+
+        with self.assertRaises(ClientError):
+            future.result()
+        # Make sure the actual file and the temporary do not exist
+        # by globbing for the file and any of its extensions
+        possible_matches = glob.glob('%s*' % self.filename)
+        self.assertEqual(possible_matches, [])
+
+    def test_download_with_nonexistent_directory(self):
+        for stubbed_response in self.stubbed_responses:
```
jamesls: Can we put a more specific name here?  Ideally someone can just read the test body to understand what the test is doing:
```
self.respond_with_single_download() # or whatever you want to call it
...
future = self.manager.download(...)
```

jamesls:
Just had a few minor comments.  The two biggest things for me:

* Seeing how we can improve the readability of the tests
* Having max_concurrency IO writer threads.  This is different from what the CLI does so I'm interested to see what if any issues there might be.  Especially given a customer can set this to something really high.

s3transfer/download.py
kyleknap: Yep that is correct. I can rename the ``IOWriterTask`` to make it more clear if that helps.

s3transfer/download.py
kyleknap: I do not think they are being handled. I pretty much copied a lot of this logic over from boto3, and it looks like it is an issue in boto3 as well. I think what we will have to do is call all of the callbacks with a negative amount to account for the reset. That is how we handle it in the uploads.

s3transfer/download.py
kyleknap: Makes sense

s3transfer/download.py
kyleknap: I wonder if that is a bug. I will look into it. This is where it is suppose to be cleaned up: https://github.com/boto/s3transfer/pull/5/files#diff-a12a8d2ff65118bd0968bf807b8cf8cdR261. Adding it as a failure_cleanup through the transfer context should handle it.

s3transfer/futures.py
kyleknap: If one thread came along and set an exception. Another thread that may not have been signaled that there was a failure may come along and clobber the exception with a success. We need to keep whatever the exception that was thrown.

s3transfer/futures.py
kyleknap: Not intentional. More of an oversight of what happens after a task is completed.

s3transfer/manager.py
kyleknap: Makes sense

s3transfer/manager.py
kyleknap: Yeah we have a base class for it. So should be fine.

s3transfer/utils.py
kyleknap: I am actually not sure. I just copied this utility from boto3: https://github.com/boto/boto3/blob/8e89d3d023b40ca3fb6f0424efd631f36a8f2533/boto3/s3/transfer.py#L442 comment and all. I am actually not sure how relevant the comment is now. I do not see how it can deadlock. Do you remember? It looks like you were the one to add the note in the first place.

tests/functional/test_download.py
kyleknap: Makes sense

tests/functional/test_download.py
kyleknap: Makes sense

tests/functional/test_download.py
kyleknap: ``__test__`` is a property that you can set to ensure test runners do not actually run the test. The BaseDownloadTest and its parent class have this set to ``False`` so we need to set it back to ``True``. I set it to ``False`` originally because it allows us to write a shared test case and all you have to do is update some of the properties based on the class of tests you are trying to run. In the case for the Base test classes these properties may not completely filled out so we should not be running/relying on them directly.

tests/integration/test_download.py
kyleknap: Probably. I need to add a functional test for this.

kyleknap:
Originally, that was the cleanest way to port of all of the existing functionality from boto3 over to the s3transfer method. I think the biggest blocker from making the architecture a single IO queue and single IO thread was that we use the IO queue as a two way for communication.

Take for example, [here](https://github.com/kyleknap/s3transfer/commit/8561dc7506700e50e78cdad65d8641ecb06884c1#diff-a12a8d2ff65118bd0968bf807b8cf8cdR268) where if we encounter an exception we signal back to the producer of the queue that we encountered a problem in writing so stop sending us more data. We cannot do this anymore if share a single queue for all downloads. Instead we would have to rely on the transfer coordinator or something.

I will look to see how far I can get in making it a single queue/thread strategy and get back to you. I have a feeling that I will have to rewrite a fair amount of the code that I used from boto3, but should be doable. I am concerned as well with a possibly large amount of threads each with a queue that could have a lot of tasks in the queue.

kyleknap:
@jamesls I got the single IO thread model working. It is in the last commit. Overall, I like it more as I feel it simplifies the architecture and code a bit.

Could you take a look at that as that is probably the largest change? The rest of the changes such as in the tests and some of the retry logic will happen in subsequent commits to make it easier to go through the io executor change.

kyleknap:
@jamesls I updated with the rest of your feedback I believe. Should be good to look at again.

jamesls:
Overall I like the new design.  I have some smaller feedback about the code (coming shortly), but design-wise I think it looks good.  The only thing I'm slightly concerned about is the two-tiered task submission where completion of the queueing of the `IOWriteTasks` from the main executor triggers submission of the `IORenameTask` to the IO executor.  It has a lot of moving parts that have to happen in the right order for that to work.  We'll have to make sure we account for all the possible edge cases, but otherwise I think we should be ok.

s3transfer/download.py
```diff
@@ -0,0 +1,311 @@
...
+            transfer_coordinator=transfer_coordinator,
+            main_kwargs={
+                'f': f,
+                'final_filename': final_filename,
+                'osutil': self._osutil
+            },
+            done_callbacks=done_callbacks,
+            is_final=True
+        )
+        submit_rename_task = functools.partial(
```
jamesls: I'd prefer not to use partials in situations like this.  It usually makes the logs/tracebacks harder to follow.  In this case, the concept of submitting a task after another task finishes seems like it's worth extracting into it's own class.

s3transfer/download.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # Submit the task to download the object.
+        download_future = self._executor.submit(
+            GetObjectTask(
+                transfer_coordinator=transfer_coordinator,
+                main_kwargs={
+                    'client': self._client,
+                    'bucket': call_args.bucket,
+                    'key': call_args.key,
+                    'f': temp_f,
```
jamesls: Can we use something more description than `f`.  I'm ok with the convention of `f` for a fileobj in a local scope, but if we're passing it around as kwargs to other objects I prefer a more descriptive name like `fileobj`, etc.

s3transfer/manager.py
```diff
@@ -27,7 +28,9 @@ def __init__(self,
                  multipart_threshold=8 * MB,
                  max_concurrency=10,
                  multipart_chunksize=8 * MB,
-                 max_queue_size=0):
+                 max_queue_size=0,
```
jamesls: Given each job will be 8MB, I think we should have the default be to cap the max queue size depending on what we feel comfortable as a max memory usage.

s3transfer/utils.py
```diff
@@ -69,6 +78,20 @@ def get_callbacks(transfer_future, callback_type):
     return callbacks
 
 
+def invoke_progress_callbacks(callbacks, bytes_transferred):
+    """Calls all progress callbacks
+
+    :param callbacks: A list of progress callbacks to invoke
+    :param bytes_transferred: The number of bytes transferred. This is passed
+        to the callbacks. If no bytes were transferred the callbacks will not
+        be invoked because no progress was achieved
```
jamesls: Should we document that you can get negative bytes transferred on retries here?

tests/functional/test_download.py
jamesls: Hmm, seems like this is easy to forget given we don't actually do this in any of our other tests.   Especially if we ever refactor/move some of the tests around, I can see us not remembering to add this value and not noticing the number of tests run dropping.  Is there any better way to not have to add this for each test subclass?

tests/functional/test_download.py
```diff
@@ -0,0 +1,350 @@
...
+        self.add_head_object_response()
+        # Insert a response that will trigger a retry after one read of the
+        # stream has been made.
+        self.add_n_retryable_get_object_responses(1, num_reads=1)
+        # Add the normal responses to simulate the download proceeding
+        # as normal after the retry.
+        self.add_successful_get_object_responses()
+
+        recorder_subscriber = RecordingSubscriber()
+        # Set the streamming to a size that is smaller than the data we
```
jamesls: typo streamming

s3transfer/download.py
```diff
@@ -0,0 +1,311 @@
...
+        raise RetriesExceededError(last_exception)
+
+
+class JoinFuturesTask(Task):
+    """A task to wait for the completion of other futures
+
+    :params future_to_wait_on: A list of futures to wait on
+    """
+    def _main(self, futures_to_wait_on):
+        pass
```
jamesls: It might be nice adding a comment on how this works.  I had to read through the `Task` base class to understand why this worked by just having a noop `_main`.

jamesls:
Just had some small feedback above, I think this is just about there.  I found the tests much easier to follow.  I definitely prefer the way they are now.

s3transfer/manager.py
kyleknap: The total memory usage currently should be max_queue_size * 8 KB (which is the size of the each chunk in the queue, not 8 MB). The CLI uses 1000 for the queue size. I am probably going to use that.

jamesls: Ah sorry I misread the default.  Have you experimented with various buffer sizes?  The CLI uses 1MB for its chunk sizes. Is there a reason we should deviate from that?

kyleknap: I have not experimented. I just used around a similar ballpark of what boto3 was at (16 KB) since it shares a fair amount of the code. With the CLI though, the default could result in a max use of 1GB (1000 * 1MB ). That seems large to me. Maybe we make the default io queue 100 with 1MB chunk size?

One issue that I can think off of the bat with 1MB is that it decreases the amount of progress callbacks, giving you not as a fine-grain measurement of progress. So I can see why 16 KB was used in boto3.

What's your thoughts on good sizes? I think a smaller stream size is better due to the better callback granularity.

jamesls: I agree that smaller stream sizes are better as long as it doesn't have any perf differences from larger chunk sizes.  Given this is all IO bound, I wouldn't expect there to be any differences, but it's worth verifying.

kyleknap: I tested some large files (~100 MB) and smaller files (~ 1MB) and I did not see a difference in performance.

tests/functional/test_download.py
kyleknap: I have not found a better way. I am up for any suggestions. I also considered adding skips but seemed undesirable as it would result in a bunch of skips in the test output.

kyleknap:
@jamesls I updated based on feedback and responded to comments.

jamesls:
:shipit:, pending adding a TODO item about the `__test__` attr.

kyleknap:
I added the todo note. Merging pull request.

Issue #7, opened 2016-02-23
No tests in pypi package


Fale:
Please, add the tests in the pypi package

Issue #6, opened 2016-02-23
No license file in the Pypi package


Fale:
Please,
Add the license file to the pypi package

PR #9, opened 2016-02-23
Add MANIFEST.in

kyleknap:
Fixes https://github.com/boto/s3transfer/issues/6
Fixes https://github.com/boto/s3transfer/issues/7

This is similar to what botocore/boto3/cli does:
https://github.com/boto/botocore/blob/develop/MANIFEST.in
https://github.com/aws/aws-cli/blob/develop/MANIFEST.in
https://github.com/boto/boto3/blob/develop/MANIFEST.in

cc @jamesls @rayluo @JordonPhillips

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,4 @@
+include README.rst
+include LICENSE.txt
+include requirements-test.txt
+graft tests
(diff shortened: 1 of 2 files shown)


jamesls:
:shipit:

PR #10, opened 2016-02-24
Implement copy method for transfer manager

kyleknap:
cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/copies.py
+++ b/s3transfer/copies.py
@@ -22,9 +22,6 @@ from s3transfer.tasks import get_callbacks
 from s3transfer.utils import calculate_range_parameter
 
 
-VERSION_ID_SUFFIX = re.compile(r'\?versionId=[^\s]+$')
-
-
 class CopyTaskSubmitter(TaskSubmitter):
     EXTRA_ARGS_TO_HEAD_ARGS_MAPPING = {
         'CopySourceIfMatch': 'IfMatch',
@@ -209,23 +206,11 @@ class CopyTaskSubmitter(TaskSubmitter):
     def _get_head_object_request_from_copy_source(self, copy_source):
         if isinstance(copy_source, dict):
             return copy.copy(copy_source)
-        elif isinstance(copy_source, str):
-            head_object_request = {}
-            bucket, key = copy_source.split('/', 1)
-            head_object_request['Bucket'] = bucket
-            head_object_request['Key'] = key
-            result = VERSION_ID_SUFFIX.search(key)
-            version_id = None
-            if result is not None:
-                head_object_request['Key'] = key[:result.start()]
-                # Make sure you strip off the ?versionId= part if it is there.
-                head_object_request['VersionId'] = key[result.start()+11:]
-            return head_object_request
         else:
             raise TypeError(
                 'Expecting dictionary formatted: '
-                '{"Bucket": bucket_name, "Key": key} or '
-                'string fomatted: "bucket_name/key" but got %s or type %s.'
+                '{"Bucket": bucket_name, "Key": key} '
+                'but got %s or type %s.'
                 % (copy_source, type(copy_source))
             )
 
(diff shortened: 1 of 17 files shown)


s3transfer/copies.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def _get_head_object_request_from_copy_source(self, copy_source):
+        if isinstance(copy_source, dict):
+            return copy.copy(copy_source)
+        elif isinstance(copy_source, str):
+            head_object_request = {}
+            bucket, key = copy_source.split('/', 1)
+            head_object_request['Bucket'] = bucket
+            head_object_request['Key'] = key
+            result = VERSION_ID_SUFFIX.search(key)
```
jamesls: Why do we have to duplicate this logic from botocore?  Is there a way we can avoid this?

tests/functional/test_copy.py
```diff
@@ -0,0 +1,520 @@
...
+            'CopySource': self.copy_source,
+            'MetadataDirective': 'REPLACE'
+        }
+
+        self.add_head_object_response(expected_params=expected_head_params)
+        self.add_successful_copy_responses(
+            expected_copy_params=expected_copy_object)
+
+        call_kwargs = self.call_kwargs
+        call_kwargs['extra_args'] = self.extra_args
```
jamesls: Isn't this the same as `self.call_kwargs['extra_args'] = self.extra_args`?

tests/integration/__init__.py
```diff
@@ -18,6 +18,16 @@
 from s3transfer.manager import TransferManager
 
 
+def recursive_delete(client, bucket_name):
+    # Recursively deletes a bucket and all of its contents.
+    objects = client.get_paginator('list_objects').paginate(
+        Bucket=bucket_name)
+    for key in objects.search('Contents[].Key'):
```
jamesls: protip: Avoid the `is not None` check with:

```
for key in objects.search('Contents[].Key || `[]`'):
```

jamesls:
Looks good, only question I had was about the str or dict version of the source param.

s3transfer/copies.py
kyleknap: Yeah I can just import the regex constant from botocore. I think I did not import it directly just incase the constant ever got moved out of handlers.py. What do you think? It could possibly be an issue since we have a major version range dependency on botocore and handlers is something we have changed (i.e. move a method in handlers.py to utils.py)

tests/functional/test_copy.py
kyleknap: No. ``self.call_kwargs`` is a property method that returns a dictionary and so it needs to be stored if modified. One option I could do if you do not like this is imbed the ``self.extra_args`` into the ``self.call_kwargs``. Thoughts?

tests/integration/__init__.py
kyleknap: Cool

kyleknap:
Should be good to look at again. After our discussion, I just completely removed the ability to pass a string as a copy source as that is a leaky abstraction that should not be used.

tests/functional/test_copy.py
jamesls: I think the confusion is it looks like an attribute so it's not obvious it's creating a new dictionary each time.  Something like `.create_call_kwargs()` or something would make that more clear.

jamesls:
Looks good, I think it's good that we only allow a dict for the copy source.

Small question about the tests but otherwise looks good.

tests/functional/test_copy.py
kyleknap: Hmm I will need to update this pattern in all of the tests then. It is a common pattern for upload, download, and copy. But I can do it.

kyleknap:
@jamesls I updated all cases where a shared property was returning a mutable object by prefixing them with ``create_`` to hopefully make that more clear. Let me know what you think.

jamesls:
:shipit:  Looks good.

PR #12, opened 2016-05-19
Acceptance tests

kyleknap:
Added a list of acceptance test that should be addressed in implementation and design of transfer manager. Most of these tests were taken from the CLI and boto3. Currently, an x indicates it has been accounted for and has been tested in some fashion. Hopefully this list will eventually get moved to the docs where we can link to which tests satisfy the acceptance test.

cc @jamesls @JordonPhillips

### DIFF
--- a/ACCEPTANCE_TESTS.rst
+++ b/ACCEPTANCE_TESTS.rst
@@ -13,11 +13,21 @@ General
 * [x] Upload single multipart file
 * [x] Upload multiple nonmultipart files
 * [x] Upload multiple multipart files
-* [x] Failed/cancelled multipart upload is aborted and leaves no orphaned parts
-* [ ] Ctrl-C of any upload does not hang
+* [x] Failed/cancelled multipart upload is aborted and leaves no orphaned parts especially for:
+
+  * [x] Failure of ``UploadPart``
+  * [x] Failure of ``CompleteMultipartUpload``
+  * [x] Failure unrelated to making an API call during upload such as read failure
+
+* [ ] Ctrl-C of any upload does not hang and the wait time is ``avg(transfer_time_iter_chunk) * some_margin``
 * [ ] Upload empty file
 * [ ] Upload nonseekable nonmultipart binary stream
 * [ ] Upload nonseekable multipart binary stream
+
+
+Region
+~~~~~~
+* [ ] Provide no or incorrect region for sig4 and be able to redirect request in fewest amount of calls as possible for multipart upload.
 
 
 Validation
@@ -34,7 +44,7 @@ Validation
 
 Extra Parameters
 ~~~~~~~~~~~~~~~~
-* [ ] Upload multipart and nonmultioart file with any of the following properties:
+* [ ] Upload multipart and nonmultipart file with any of the following properties:
 
   * [x] ACL's
   * [x] CacheControl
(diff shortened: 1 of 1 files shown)


ACCEPTANCE_TESTS.rst
```diff
@@ -0,0 +1,163 @@
...
+  * [ ] Make chunksize 5 MB when the provided chunksize is less
+  * [ ] Make chunksize 5 GB when the provided chunksize is more
+  * [ ] Increase chunksize till the maximum number of parts for multipart upload is less than or equal to 10,000 parts
+
+* [ ] Before upload, ensure upload is nonmultipart if the file size is less than 5 MB no matter the provided multipart threshold.
+
+
+Extra Parameters
+~~~~~~~~~~~~~~~~
+* [ ] Upload multipart and nonmultioart file with any of the following properties:
```
jamesls: typo

ACCEPTANCE_TESTS.rst
```diff
@@ -0,0 +1,163 @@
...
+* [ ] Copy multipart and nonmultipart objects of a specific version
+* [ ] Copy multipart and nonmultipart objects using requester pays
+* [ ] Copy multipart and nonmultipart objects using a sse-c key
+* [ ] Copy multipart and nonmultipart objects using a copy source sse-c key
+* [ ] Copy multipart and nonmultipart objects using a copy source sse-c key and sse-c key
+
+
+Cross-Bucket
+~~~~~~~~~~~~
+* [ ] Copy single nonmultipart object across sigv4 regions
```
jamesls: Is it worth calling out cases where we don't know (or aren't given) the region and we have to do region discovery?

ACCEPTANCE_TESTS.rst
```diff
@@ -0,0 +1,163 @@
...
+------------
+
+General
+~~~~~~~
+* [x] Upload single nonmultipart file
+* [x] Upload single multipart file
+* [x] Upload multiple nonmultipart files
+* [x] Upload multiple multipart files
+* [x] Failed/cancelled multipart upload is aborted and leaves no orphaned parts
+* [ ] Ctrl-C of any upload does not hang
```
jamesls: In general I agree. It would be nice to have a even more aggressive time than the CLI has in its tests.  Right now, I think the CLI allows up to 30 seconds to exit.

ACCEPTANCE_TESTS.rst
```diff
@@ -0,0 +1,163 @@
...
+Upload Tests
+------------
+
+General
+~~~~~~~
+* [x] Upload single nonmultipart file
+* [x] Upload single multipart file
+* [x] Upload multiple nonmultipart files
+* [x] Upload multiple multipart files
+* [x] Failed/cancelled multipart upload is aborted and leaves no orphaned parts
```
jamesls: Is it worth calling out a few sub conditions?

* Any failing upload part results in aborted upload.
* The complete multipart upload failing will result in an aborted upload.
* Not being able to read from the source file (or some other error unrelated to making an API call) results in an aborted upload.

ACCEPTANCE_TESTS.rst
kyleknap: We don't do this now. Right? But planning to add it in botocore. So I guess my question is should we be documenting behavior implemented in botocore here in this document?

ACCEPTANCE_TESTS.rst
kyleknap: How many seconds were you thinking? Does less than 10 seconds sound good? Ideally it would be near instantly of course.

ACCEPTANCE_TESTS.rst
kyleknap: Not opposed.

ACCEPTANCE_TESTS.rst
jamesls: Not sure about an absolute value, but I would expect it would be `avg(transfer_time_iter_chunk) * margin`.  I think if we're using 8k buffer sizes for downloads, a couple of seconds should be sufficient.

jamesls:
:shipit:  I think this looks like a good start.

ACCEPTANCE_TESTS.rst
kyleknap: I am good with that

ACCEPTANCE_TESTS.rst
kyleknap: Now that I think of it the multithreading aspect of it may be tricky in region detection. So I will add a note for all transfers.

jamesls:
:shipit:

PR #13, opened 2016-05-23
Wrap task submission for transfer into tasks

kyleknap:
This is a pretty large refactoring of the architecture, but I think that it is worth it and necessary in the long run as we add features. In short, the change wraps the ``TaskSubmitter`` logic that submits all of the tasks to execute a transfer into a ``SubmissionTask`` itself and these submission tasks are then executed by an executor specific for running submission tasks. This is really helpful for the following reasons:

* The submission of tasks are no longer handled in the main thread, which makes it quicker to enqueue transfer request and get a transfer future back. For example, in downloads and copies you are no longer doing a ``HeadObject ``in the main thread. This is also going to be needed for streaming uploads (i.e. we do not want to stream the entire file in the main thread).
* Exception handling is more robust such that if there was an error during submission before the final task was submitted, bad things could have happened such as hanging or cleanup tasks not being ran. This is no longer the case.
* ``on_queued`` callbacks happen in the submission thread pool. This allows you to make slower calls in these callbacks without slowing down your speed in adding more transfer requests. This will be critical if you want to try to do complete copies for multipart which could require multiple API requests.

As to the changes that were made in the refactoring, they include:

* Remove the ``TaskSubmitter`` class in place of a ``SubmissionTask`` class. They relatively have the same functionality but the ``SubmissionTask`` subclasses from the ``Task`` class so it can be ran by an executor.
* Added a new executor to complete ``SubmissionTask`` that does not interfere with the main executor.
* Moved the running of failure cleanups and done callbacks into the ``TransferCoordinator``. This was changed because logically it made sense that coordinator runs the callbacks associated to it when it announces done especially since it is no longer the responsibility of a single know task to announce that the transfer is done (This can happen in the submission task if it fails or the designated final task).

Let me know if you have any questions. I tried my best to add comments and documentation at the trickier parts of the code.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/manager.py
+++ b/s3transfer/manager.py
@@ -11,14 +11,16 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from s3transfer.utils import unique_id
+from s3transfer.utils import get_callbacks
 from s3transfer.utils import disable_upload_callbacks
 from s3transfer.utils import enable_upload_callbacks
 from s3transfer.utils import CallArgs
 from s3transfer.utils import OSUtils
 from s3transfer.futures import BoundedExecutor
-from s3transfer.download import DownloadTaskSubmitter
-from s3transfer.upload import UploadTaskSubmitter
-from s3transfer.copies import CopyTaskSubmitter
+from s3transfer.futures import get_transfer_future_with_components
+from s3transfer.download import DownloadSubmissionTask
+from s3transfer.upload import UploadSubmissionTask
+from s3transfer.copies import CopySubmissionTask
 
 
 MB = 1024 * 1024
@@ -27,9 +29,11 @@ MB = 1024 * 1024
 class TransferConfig(object):
     def __init__(self,
                  multipart_threshold=8 * MB,
-                 max_concurrency=10,
                  multipart_chunksize=8 * MB,
-                 max_queue_size=0,
+                 max_request_concurrency=10,
+                 max_submission_concurrency=5,
+                 max_request_queue_size=0,
+                 max_submission_queue_size=0,
                  max_io_queue_size=1000,
                  num_download_attempts=5):
         """Configurations for the transfer mangager
@@ -37,15 +41,27 @@ class TransferConfig(object):
         :param multipart_threshold: The threshold for which multipart
             transfers occur.
 
-        :param max_concurrency: The maximum number or requests that
-            can happen at a time.
+        :param max_request_concurrency: The maximum number of S3 API
+            transfer-related requests that can happen at a time.
+
+        :param max_submission_concurrency: The maximum number of threads
+            processing a call to a TransferManager method. Processing a
+            call usually entails determining which S3 API requests that need
+            to be enqueued, but does **not** entail making any of the
+            S3 API data transfering requests needed to perform the transfer.
+            The threads controlled by ``max_request_concurrency`` is
+            responsible for that.
 
         :param multipart_chunksize: The size of each transfer if a request
             becomes a multipart transfer.
 
-        :param max_queue_size: The maximum amount of requests that
-            can be queued at a time. A value of zero means that there
+        :param max_request_queue_size: The maximum amount of S3 API requests
+            that can be queued at a time. A value of zero means that there
             is no maximum.
+
+        :param max_submission_queue_size: The maximum amount of
+            TransferManager method calls that can be queued at a time. A value
+            of zero means that there is no maximum.
 
         :param max_io_queue_size: The maximum amount of read parts that
             can be queued to be written to disk per download. A value of zero
@@ -63,9 +79,11 @@ class TransferConfig(object):
             number of exceptions retried by botocore.
         """
         self.multipart_threshold = multipart_threshold
-        self.max_concurrency = max_concurrency
         self.multipart_chunksize = multipart_chunksize
-        self.max_queue_size = max_queue_size
+        self.max_request_concurrency = max_request_concurrency
+        self.max_submission_concurrency = max_submission_concurrency
+        self.max_request_queue_size = max_request_queue_size
+        self.max_submission_queue_size = max_submission_queue_size
         self.max_io_queue_size = max_io_queue_size
         self.num_download_attempts = num_download_attempts
 
@@ -123,10 +141,20 @@ class TransferManager(object):
         if config is None:
             self._config = TransferConfig()
         self._osutil = OSUtils()
-        self._executor = BoundedExecutor(
-            max_size=self._config.max_queue_size,
-            max_num_threads=self._config.max_concurrency
+
+        # The executor responsible for making S3 API transfer requests
+        self._request_executor = BoundedExecutor(
+            max_size=self._config.max_request_queue_size,
+            max_num_threads=self._config.max_request_concurrency
         )
+
+        # The executor responsible for submitting the necessary tasks to
+        # perform the desired transfer
+        self._submission_executor = BoundedExecutor(
+            max_size=self._config.max_submission_queue_size,
+            max_num_threads=self._config.max_submission_concurrency
+        )
+
         # There is one thread available for writing to disk. It will handle
         # downloads for all files.
         self._io_executor = BoundedExecutor(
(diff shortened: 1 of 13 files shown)


s3transfer/manager.py
```diff
@@ -27,26 +29,40 @@
 class TransferConfig(object):
     def __init__(self,
                  multipart_threshold=8 * MB,
-                 max_concurrency=10,
                  multipart_chunksize=8 * MB,
-                 max_queue_size=0,
+                 max_request_concurrency=10,
```
jamesls: Given this a backwards incompatible change, what's the plan to interop with the boto3 transfer config object?

s3transfer/manager.py
```diff
@@ -273,6 +293,46 @@ def _validate_all_known_args(self, actual, allowed):
                     "must be one of: %s" % (
                         kwarg, ', '.join(allowed)))
 
+    def _submit_transfer(self, call_args, submission_task_cls,
+                         extra_main_kwargs=None):
+        if not extra_main_kwargs:
+            extra_main_kwargs = {}
+
+        # Create a TransferFuture to return back to the user
+        transfer_future, components = get_transfer_future_with_components(
```
jamesls: Is there a reason this needs to live external to this class?  Seems like it belongs in this class.

s3transfer/tasks.py
```diff
@@ -245,6 +157,97 @@ def _get_all_main_kwargs(self):
...
+            to the _submit() method
+        """
+        try:
+            # Before submitting any tasks, run all of the on_queued callbacks
+            on_queued_callbacks = get_callbacks(transfer_future, 'queued')
+            for on_queued_callback in on_queued_callbacks:
+                on_queued_callback()
+
+            # Once callbacks have been ran set the status to running.
+            self._transfer_coordinator.set_status_to_running()
```
jamesls: Is this right?  The transfer coordinator's status of running is for the actual transfer task right (i.e the GetObject() from S3).  Shouldn't the actual transfer task be the one that flips the status to running?

jamesls:
Overall pretty straightforward.  Two general pieces of feedback:

1. It would be helpful to add `.debug()` logs to the various task submission/queueing parts.  There's likely to be tricky concurrency issues, and the only insight we generally get is by asking for debug logs.  Being able to see how things are getting queued/processed will make this easier to troubleshoot.
2. Getting to the bottom of the perf discrepancy between the develop branch as well as `aws s3 cp`.  Would also be worth checking if this is only affecting downloads or if uploads is affected as well.

s3transfer/manager.py
kyleknap: I was planning to have a compatibility layer anyways that will do some conversion such that if they gave me the deprecated Config object, then I would convert it to the appropriate variable. I have already made a need to add this compatibility layer anyways with the way subscribers are provided (i.e. they have to be of a certain class abstraction and you can provide more than one subscriber).

I feel like this name is much more clear on what this actually tunes given that I have three executors/pools of threads and was worth the change. What do you think?

s3transfer/manager.py
kyleknap: Not really. I can move it to this class. There is not really a reason it needs to be external other than it is a little easier to test, but should not be an issue.

s3transfer/manager.py
jamesls: To clarify I was asking specifically how you plan on mapping these.  Is max_concurrency -> max_request_concurrency?  Are you not going to allow users to specify max submission concurrency in the existing boto3 transfer config?

kyleknap: I was not going to let them tune stuff like submission queue size and number of threads it to start. It does not make sense especially since the TransferManager is already blocking and I need to keep that backwards compatibility for that as well. So something roughly like this:
```py
from s3transfer.manager import TransferConfig as S3TransferConfig


class TransferConfig(S3TransferConfig):
    ALIAS = {
        'max_concurrency': 'max_request_concurrency'
    }
    def __init__(self,
                 multipart_threshold=8 * MB,
                 max_concurrency=10,
                 multipart_chunksize=8 * MB,
                 num_download_attempts=5,
                 max_io_queue=100):
         super(TransferConfig, self).__init__(... # Account for mappings as well)
        for alias in self.ALIAS:
            setattr(self, alias, getattr(self, self.ALIAS[alias])
```
If we want the option for the manager to be non-blocking then we would look into adding the extra tuning variables. What do you think?

s3transfer/tasks.py
kyleknap: So I guess this depends on the definition of ``running``. Should it be when the first transfer API request is made like ``GetObject`` or ``PutObject`` or should it be once work has started for the request? I saw it as the later where if I called ``HeadObject`` or pulled data from the stream for an upload then my transfer has started and thus I put it in the "running" state. What do you think?

kyleknap:
I will add more debug logs. That makes sense.

As to the performance comparison, I will put some numbers up, but I am not seeing too big of a difference between this PR and the develop branch. So I think addressing it in a seperate PR would be more appropriate. It must be as fast as or faster than the CLI before we start migrating it.

s3transfer/tasks.py
kyleknap: I talked to you about this and I looked back through the code and it looks like we do not even expose the ``status`` property on a ``Transferfuture`` (we only use the status internally) so a user could not even probe into what the ``status`` is of a future to report a discrepancy. What is your thoughts on tabling this to when we expose ``status`` on the future because we will have a better idea of what each state should mean given as we would only expose it if there is enough reason to make it a public feature?

kyleknap:
@jamesls I tested the uploads out and they are fine. They are actually faster than the CLI.

The downloads I am still testing. I found one issue that gets the develop branch back to the speed of the CLI, where our default configuration is not inappropriate for downloads.

Consider the case where we have a fast network and are bounded more by io. Given that our default io queue size is a thousand and the default io chunk size is 8KB, then the most amount of data that can be stored in the queue is around 8 MB. However, this is a problem because each downloaded part is 8 MB as well so if the threads are downloading it really quick there is a bottleneck in the queue where the queue may fill up because the enqueuing is much quicker than the io thread is processing them. So I think we need to bump up the value of the io chunksize or io queue size to something closer to the value of (max_request_threads * request_chunksize). This makes sense too why the CLI is faster because for the CLI the max io queue size is 1000 while the io chunks are 1MB.

What is your thoughts on this? Should we bump up the default io queue size of io chunksize? After testing bumping the io chunksize to 64 KB is the best, but I see similar results by increasing the io queue size. So I would be inclined to pick the bumping of chunksize even though we may lose some granularity in the progress callbacks.

However, with this change my PR is still twice as slow as the CLI, but it is definitely faster than before. I need to do more digging on that.

Also on a side note, I commented on two other comments you had.

kyleknap:
Hmm, after testing changing the io chunksize to 64 KB, it is actually as fast as the CLI now, I may have just not been using the correct io chunk size in the first place.

kyleknap:
@jamesls I addressed all comments by updating the code or by responding back to the comment. Should be good to look at again.

s3transfer/manager.py
jamesls: Seems reasonable.

We'll probably need to loop back and work out good class names for the config classes to make sure no one confuses the similarly named transfer config classes we have.

s3transfer/manager.py
jamesls: Preferred way I like to handle this is via dep. injection if you want it to be separately tested.  I always find it a bit odd to have internal, encapsulated instance methods that essentially reach into the global namespace to leverage utility functions.  Sometimes it makes sense, but I find it easier if that dependency is made explicit in the `__init__`.

s3transfer/tasks.py
jamesls: Seems fine.  I'd also say that we should document/comment on the various states of the transfer future and make it clear what our definition of `running` is, if no other reason to make it easier for us to understand the code.

s3transfer/manager.py
kyleknap: Oh I made the ``get_transfer_future_with_components`` internal to the class --> ``_get_transfer_future_with_components`` so that may just be an old comment. I agree there was not a reason it needed to be in ``futures.py`` given it was only used in the manager.

s3transfer/tasks.py
kyleknap: Makes sense I can better document it then.

jamesls:
Ah, your explanation makes a lot of sense, good deep dive on figuring that out.  I think the math works out nicely as well with the defaults:

* 1k max queue * 64kb IO chunks = 62.5MB max mem allocated to IO queue
* 8MB part sizes with 64kb UI chunks = 128 parts per thread = 1280 possible IO slots needed before threads can move on to the next part.  I'd also be cool with bumping the default queue to 1280, but I don't know if it makes a noticeable difference because that's a worst case scenario.

Looking over the rest of the changes now.

jamesls:
:shipit:  Looks good to me.

kyleknap:
I updated the status documentation for running as well. Merging.

PR #14, opened 2016-06-06
Add support for seekable file objects

kyleknap:
First, this pull request is based off of this PR: https://github.com/boto/s3transfer/pull/13. So you probably would want to review that one first.

As to the changes made, in this PR they are as follows:
* Cleaned up functional tests for uploads. Mainly making them easier to read.
* Abstracted logic for different source inputs for uploads into different classes to make it much easier to add new strategies for consuming different sources of data.
* Add support for seekable files.

Note that right now there is no protection from having too many chunks in memory. I am looking to add this in a follow up pull request to make it less to review at a time.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/utils.py
+++ b/s3transfer/utils.py
@@ -221,7 +221,8 @@ class DeferredOpenFile(object):
         return self._fileobj.tell()
 
     def close(self):
-        self._fileobj.close()
+        if self._fileobj:
+            self._fileobj.close()
 
     def __enter__(self):
         self._open_if_needed()
--- a/s3transfer/utils.py
+++ b/s3transfer/utils.py
@@ -179,8 +179,60 @@ class OSUtils(object):
         rename_file(current_filename, new_filename)
 
 
+class DeferredOpenFile(object):
+    OPEN_METHOD = open
+
+    def __init__(self, filename, start_byte=0):
+        """A class that defers the opening of a file till needed
+
+        This is useful for deffering opening of a file till it is needed
+        in a separate thread, as there is a limit of how many open files
+        there can be in a single thread for most operating systems. The
+        file gets opened in the following methods: ``read()``, ``seek()``,
+        and ``__enter__()``
+
+        :type filename: str
+        :param filename: The name of the file to open
+
+        :type start_byte: int
+        :param start_byte: The byte to seek to when the file is opened.
+        """
+        self._filename = filename
+        self._fileobj = None
+        self._start_byte = start_byte
+
+    def _open_if_needed(self):
+        if self._fileobj is None:
+            self._fileobj = self.OPEN_METHOD(self._filename, 'rb')
+            if self._start_byte != 0:
+                self._fileobj.seek(self._start_byte)
+
+    def read(self, amount=None):
+        self._open_if_needed()
+        return self._fileobj.read(amount)
+
+    def seek(self, where):
+        self._open_if_needed()
+        self._fileobj.seek(where)
+
+    def tell(self):
+        if self._fileobj is None:
+            return self._start_byte
+        return self._fileobj.tell()
+
+    def close(self):
+        self._fileobj.close()
+
+    def __enter__(self):
+        self._open_if_needed()
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        self.close()
+
+
 class ReadFileChunk(object):
-    def __init__(self, fileobj, start_byte, chunk_size, full_file_size,
+    def __init__(self, fileobj, chunk_size, full_file_size,
                  callbacks=None, enable_callbacks=True):
         """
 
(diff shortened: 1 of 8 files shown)


kyleknap:
Here is some testing as well with my machine

**Using filename and s3transfer:**
```py
import time

import boto3

c = boto3.client('s3')
key = '100MB'

manager = TransferManager(c)
# boto3.set_stream_logger('s3transfer')
start_time = time.time()
future = manager.upload(key, 'mybucketfoo', key)
queue_time = time.time()
print("Time to queue upload: %s" % (queue_time - start_time))
future.result()
end_time = time.time()
print("Time to complete upload: %s" % (end_time - start_time))
```
**Output:**
```
Time to queue upload: 0.000439167022705
Time to complete upload: 3.91357517242
```
**Using fileobj and s3transfer:**
```py
import time

import boto3

c = boto3.client('s3')
key = '100MB'

manager = TransferManager(c)
# boto3.set_stream_logger('s3transfer')
start_time = time.time()
with open(key, 'rb') as f:
    future = manager.upload(key, 'mybucketfoo', key)
    queue_time = time.time()
    print("Time to queue upload: %s" % (queue_time - start_time))
    future.result()
    end_time = time.time()
    print("Time to complete upload: %s" % (end_time - start_time))
```
**Output:**
```
Time to queue upload: 0.000407934188843
Time to complete upload: 3.91228914261
```

And then comparing this to the CLI for completeness:
```
$ time aws s3 cp 100MB s3://mybucketfoo/
upload: ./100MB to s3://mybucketfoo/100MB          

real	0m6.562s
user	0m1.112s
sys	0m0.615s

```
And stream (note the multipart case nonseekable is very similar to s3transfer multipart seekable case):
```
$ cat 100MB | time aws s3 cp - s3://mybucketfoo/
        5.35 real         1.04 user         0.64 sys
```
So it looks like in general this is significantly faster than the CLI which is good.

tests/functional/test_upload.py
```diff
@@ -99,20 +99,53 @@ def create_stubbed_responses(self):
...
+        upload_responses = stubbed_responses[0:1]
+        if len(stubbed_responses) > 1:
+            upload_responses = stubbed_responses[1:-1]
+
+        # Add the expected create multipart upload params.
+        if expected_create_mpu_params:
+            stubbed_responses[0][
+                'expected_params'] = expected_create_mpu_params
+
+        # Add any expected copy parameters.
```
kyleknap: copy --> upload

tests/functional/test_upload.py
```diff
@@ -175,75 +209,72 @@ def create_expected_progress_callback_info(self):
...
+        expected_upload_params = []
+        # Add expected parameters to the copy part
+        num_parts = 3
+        for i in range(num_parts):
+            expected_upload_params.append(
+                {
+                    'Bucket': self.bucket,
+                    'Key': self.key,
+                    'UploadId': self.multipart_id,
+                    'Body': mock.ANY,
```
kyleknap: use stub.ANY instead of mock.ANY

s3transfer/utils.py
```diff
@@ -179,8 +179,60 @@ def rename_file(self, current_filename, new_filename):
         rename_file(current_filename, new_filename)
 
 
+class DeferredOpenFile(object):
+    OPEN_METHOD = open
```
kyleknap: Flagging for opinions on use of class variable that allows you to control how the file is opened. I can see being able to switch how you open the file as being useful, but is not getting exercised. Really only switch it out in the tests. Without a way of injecting an open method there is not a great way to test it other than patching, which I think is the least desirable option.

s3transfer/utils.py
```diff
@@ -179,8 +179,60 @@ def rename_file(self, current_filename, new_filename):
...
+
+    def __exit__(self, *args, **kwargs):
+        self.close()
+
+
 class ReadFileChunk(object):
-    def __init__(self, fileobj, start_byte, chunk_size, full_file_size,
+    def __init__(self, fileobj, chunk_size, full_file_size,
                  callbacks=None, enable_callbacks=True):
         """
```
kyleknap: Need to codeblock this. Probably will copy and paste from boto3.

kyleknap: Now addressed.

s3transfer/utils.py
kyleknap: Decided to keep this as it is useful to override how the file is opened.

kyleknap:
@jamesls 
It should be good to look at again. In the recent changes, I did the following:

* make ``get_upload_input_manager_cls`` internal to ``UploadSubmissionTask``
* exposed an ``open_file_chunk_reader_from_fileobj`` on ``OSUtil`` and decoupled ``DeferredFileOpen`` from ``ReadFileChunk``.
* Cleaned up the upload functional tests to make it more readable and explicit on what I am trying to test by making the assertions and helper methods more understandable from a high-level perspective.

Let me know what you think or if there is anything I missed.

tests/unit/test_utils.py
```diff
@@ -181,6 +194,55 @@ def test_rename_file(self):
...
+        self.assertEqual(content, self.contents[2:4])
+        self.assertEqual(self.open_called_count, 1)
+
+    def test_tell(self):
+        self.deffered_open_file.tell()
+        # tell() should not have opened the file if it has not been seeked
+        # or read because we know the start bytes upfront.
+        self.assertEqual(self.open_called_count, 0)
+
+        self.deffered_open_file.seek(2)
```
jamesls: type, deferred.

jamesls:
:shipit:  Looks good to me.

PR #15, opened 2016-06-09
Fix issue in tests

kyleknap:
The issue is that since we did not call result on the future, the process still
maybe going during the tearDown and as a result on systems like windows we
may be trying to remove the temporary file for downloads while the transfer
manager is still acting on it, which causes issues when two processes are
performing on the same file.

cc @jamesls @JordonPhillips


kyleknap:
Cool. Regression in stubber fixed. It should be good to look at.

jamesls:
:shipit:

PR #17, opened 2016-06-15
Add ability to limit number of in-memory upload chunks

kyleknap:
This is currently helpful for multipart seekable files where we have to store
the data into memory. This will allow users to limit the number of these
partitioned chunks of data can be in memory at a time. This will also be needed
for nonseekable uploads and possibly nonseekable downloads.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -297,12 +297,12 @@ class BoundedExecutor(object):
         """
         self._max_num_threads = max_num_threads
         self._executor = self.EXECUTOR_CLS(max_workers=self._max_num_threads)
-        self._tags_to_track = ['all']
+        self._tags_to_track = [ALL_TAG]
         self._max_sizes = {
-            'all': max_size
+            ALL_TAG: max_size
         }
         self._currently_running_futures = {
-            'all': set()
+            ALL_TAG: set()
         }
         if tag_max_sizes:
             for tag, max_size in tag_max_sizes.items():
@@ -339,10 +339,10 @@ class BoundedExecutor(object):
     def _get_tags_to_operate_on(self, tag):
         # No matter the tag provided, each future is included under the 'all'
         # tag to represent its membership in the set of all running futures.
-        tag_to_operate_on = ['all']
+        tag_to_operate_on = [ALL_TAG]
         # Only add the tag to the set of the tags to operate on if the tag
         # is not equal to all and the tag is being tracked for max sizes.
-        if tag != 'all' and tag in self._tags_to_track:
+        if tag is not ALL_TAG and tag in self._tags_to_track:
             tag_to_operate_on.append(tag)
         return tag_to_operate_on
 
@@ -372,3 +372,12 @@ class BoundedExecutor(object):
             # track of how many futures are currently being ran.
             if self._max_sizes[tag] != 0:
                 self._currently_running_futures[tag].add(future)
+
+
+class FutureTag(object):
+    def __init__(self, name):
+        self.name = name
+
+
+ALL_TAG = FutureTag('all')
+IN_MEMORY_UPLOAD_TAG = FutureTag('in_memory_upload')
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from concurrent import futures
+from collections import namedtuple
 import copy
 import logging
 import threading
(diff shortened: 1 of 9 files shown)


s3transfer/manager.py
```diff
@@ -79,6 +80,16 @@ def __init__(self,
...
             number of exceptions retried by botocore.
+
+        :param max_in_memory_upload_chunks: The number of chunks that can
+            be stored in memory at a time for all ongoing uploads. This
+            pertains to chunks of data that need to be stored in memory
+            during an upload if the data is sourced from a file-like object.
+            The total expected memory footprint due to in-memory upload
+            chunks is roughly equal to:
+
+                max_in_memory_upload_chunks * multipart_chunksize
```
kyleknap: Looking at the code again, it actually appears to be that the calculation to determine how many in-flight upload chunks should be ``max_in_memory_upload_chunks * multipart_chunksize + submission_concurrency * multipart_chunksize`` because even though the submission gets blocked by the executor, the submission thread will still call the next iteration and that in-memory chunk will be blocked in getting submitted to request executor. Does this make sense?

I wonder if there is a better abstraction I can do to handle this. The hard part is that I need to communicate the number of in-flight chunks between two thread pools especially since you cannot make the assumption that a specific type of input will produce in-flight chunks (you need to know if it is multipart or not). I wonder if I can just do a better job of naming the tuning parameter and updating the description?

kyleknap:
@jamesls 
I updated the code based on feedback that both you and I noticed:

* Use a constant/enum object for specific tags for consistent coordination
* I updated the documentation to give better information about maximum in-memory streaming chunks.

Should be good to look at again. Let me know what you think.

s3transfer/futures.py
```diff
@@ -285,44 +285,99 @@ def __init__(self, max_size, max_num_threads):
...
             # then we do not need to store the future as we are not keeping
             # track of how many futures are currently being ran.
-            if self._max_size is not None:
-                self._currently_running_futures.add(future)
-        return future
+            if self._max_sizes[tag] != 0:
+                self._currently_running_futures[tag].add(future)
+
+
+class FutureTag(object):
```
jamesls: Did you intend for object identity to be a requirement for uniqueness?  Otherwise, a namedtuple would be good here.  You'd get equality and good reprs() out of the box.

tests/functional/test_upload.py
```diff
@@ -275,6 +275,31 @@ def test_upload_for_fileobj(self):
         self.assert_expected_client_calls_were_correct()
         self.assert_upload_part_bodies_were_correct()
 
+    def test_limits_in_memory_chunks_for_fileobj(self):
+        # Limit the maximum in memory chunks to one but make number of
+        # threads more than one. This means that the upload will have to
+        # happen sequentially despite having many threads available because
+        # data is sequentially partitioned into chunks in memory and since
+        # there can only every be one in memory chunk, each upload part will
+        # have to happen one at a time.
```
jamesls: I get the explanation, but I'm having trouble following exactly how you're testing that upload parts happened sequentially in the functional test.  All I see is `assert_expected_client_calls_were_correct` and `assert_upload_part_bodies_were_correct`.

jamesls:
I'm still concerned about the complexity of this approach, but don't have anything better to suggest right now.  I think we can try it out and see how it goes.

Just had a couple of small questions above.

s3transfer/futures.py
kyleknap: It was intended, but I am not too attached to this requirement. I can change it to a named tuple, if it makes it easier to work with/read.

jamesls: Or at least add a repr then?  That way printing the dictionary would make it easier to debug.

tests/functional/test_upload.py
jamesls: Let me phrase it another way.  Where in this test method is that expectation called out?  That "if there is a discrepancy then we'll fail in the stubber."  Is that implicit in the sequencing of calls to `self.add_*` below?

If so, then I think that's pretty reasonable.

kyleknap: Yeah it is implicit to the ``self.add_*`` methods of the stubber. I will add comments to make that more clear about it implicitly having the assertion of order.

s3transfer/futures.py
kyleknap: I ended up making it a namedtuple. It required less code and a brand new class was not needed.

kyleknap:
@jamesls I addressed your comments I think. Let me know what you think.

jamesls:
:shipit:  Looks good to me.

PR #18, opened 2016-06-21
Add support for downloading to seekable files

kyleknap:
This pull request does the following:
* Add a DownloadOutputManager that gives us flexibility in a similar fashion as the UploadInputManager
* Add support for seekable files.

Here is some comparisons:
**Seekable file download**
**Code:**
```py
import time

import boto3

c = boto3.client('s3')
key = '100MB'

start_time = time.time()
manager = TransferManager(c)
with open(key, 'wb') as f:
    future = manager.download('mybucketfoo', key, f)
    queue_time = time.time()
    print("Time to queue download: %s" % (queue_time - start_time))
    future.result()
    end_time = time.time()
    print("Time to complete download: %s" % (end_time - start_time))
```
**Results:**
```
Time to queue download: 0.00941491127014
Time to complete download: 3.51214289665
```

**Filename**
**Code**
```py
import time

import boto3

c = boto3.client('s3')
key = '100MB'

start_time = time.time()
manager = TransferManager(c)
future = manager.download('mybucketfoo', key, key)
queue_time = time.time()
print("Time to queue download: %s" % (queue_time - start_time))
future.result()
end_time = time.time()
print("Time to complete download: %s" % (end_time - start_time))
```
**Results**
```
Time to queue download: 0.00508308410645
Time to complete download: 3.74164819717
```

**CLI Comparison**
```
$ time aws s3 cp s3://mybucketfoo/100MB 100MB
download: s3://mybucketfoo/100MB to ./100MB        

real	0m3.738s
user	0m0.844s
sys	0m0.491s
```
cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/compat.py
+++ b/s3transfer/compat.py
@@ -50,3 +50,27 @@ else:
         return inspect.getargspec(func)[2]
 
     SOCKET_ERROR = socket.error
+
+
+def seekable(fileobj):
+    """Backwards compat function to determine if a fileobj is seekable
+
+    :param fileobj: The file-like object to determine if seekable
+
+    :returns: True, if seekable. False, otherwise.
+    """
+    try:
+        # If the fileobj has a seekable attr, try calling the seekable()
+        # method on it.
+        if hasattr(fileobj, 'seekable'):
+            return fileobj.seekable()
+        # If there is no seekable attr, check if the object can be seeked
+        # or telled. If it can, try to seek to the current position.
+        elif hasattr(fileobj, 'seek') and hasattr(fileobj, 'tell'):
+            fileobj.seek(0, 1)
+            return True
+        # Else, the fileobj is not seekable
+        return False
+    except (OSError, IOError):
+        # If an io related error was thrown as well then it is not seekable.
+        return False
--- a/s3transfer/compat.py
+++ b/s3transfer/compat.py
@@ -59,18 +59,18 @@ def seekable(fileobj):
 
     :returns: True, if seekable. False, otherwise.
     """
-    try:
-        # If the fileobj has a seekable attr, try calling the seekable()
-        # method on it.
-        if hasattr(fileobj, 'seekable'):
-            return fileobj.seekable()
-        # If there is no seekable attr, check if the object can be seeked
-        # or telled. If it can, try to seek to the current position.
-        elif hasattr(fileobj, 'seek') and hasattr(fileobj, 'tell'):
+    # If the fileobj has a seekable attr, try calling the seekable()
+    # method on it.
+    if hasattr(fileobj, 'seekable'):
+        return fileobj.seekable()
+    # If there is no seekable attr, check if the object can be seeked
+    # or telled. If it can, try to seek to the current position.
+    elif hasattr(fileobj, 'seek') and hasattr(fileobj, 'tell'):
+        try:
             fileobj.seek(0, 1)
             return True
-        # Else, the fileobj is not seekable
-        return False
-    except (OSError, IOError):
-        # If an io related error was thrown as well then it is not seekable.
-        return False
+        except (OSError, IOError):
+            # If an io related error was thrown then it is not seekable.
+            return False
+    # Else, the fileobj is not seekable
+    return False
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -131,6 +131,27 @@ class DownloadFilenameOutputManager(DownloadOutputManager):
         return f
 
 
+class DownloadSeekableOutputManager(DownloadOutputManager):
+    @classmethod
+    def is_compatible(self, download_target):
+        return (
+            hasattr(download_target, 'seek') and
+            hasattr(download_target, 'tell')
+        )
+
+    def get_fileobj_for_io_writes(self, transfer_future):
+        # Return the fileobj provided to the future.
+        return transfer_future.meta.call_args.fileobj
+
+    def get_final_io_task(self):
+        # This task will serve the purpose of signaling when all of the io
+        # writes have finished so done callbacks can be called.
+        return CompleteDownloadNOOPTask(
+            transfer_coordinator=self._transfer_coordinator,
+            is_final=True
+        )
+
+
 class DownloadSubmissionTask(SubmissionTask):
     """Task for submitting tasks to execute a download"""
 
@@ -146,6 +167,7 @@ class DownloadSubmissionTask(SubmissionTask):
         """
         download_manager_resolver_chain = [
             DownloadFilenameOutputManager,
+            DownloadSeekableOutputManager
         ]
 
         fileobj = transfer_future.meta.call_args.fileobj
@@ -431,3 +453,9 @@ class IORenameFileTask(Task):
     def _main(self, fileobj, final_filename, osutil):
         fileobj.close()
         osutil.rename_file(fileobj.name, final_filename)
+
+
+class CompleteDownloadNOOPTask(Task):
+    """A NOOP task to serve as an indicator that the download is complete"""
+    def _main(self):
+        pass
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -82,6 +82,11 @@ class DownloadOutputManager(object):
     def get_final_io_task(self):
         """Get the final io task to complete the download
 
+        This is needed because based on the architecture of the TransferManager
+        the final tasks will be sent to the IO executor, but the executor
+        needs a final task for it to signal that the transfer is done and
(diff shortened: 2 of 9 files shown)


s3transfer/download.py
```diff
@@ -38,9 +39,123 @@
...
+    def __init__(self, osutil, transfer_coordinator):
+        super(DownloadFilenameOutputManager, self).__init__(
+            osutil, transfer_coordinator)
+        self._final_filename = None
+        self._temp_filename = None
+        self._temp_fileobj = None
+
+    @classmethod
+    def is_compatible(self, download_target):
+        return isinstance(download_target, six.string_types)
```
JordonPhillips: Maybe you should check here to see if you have write access to the target and validate that it's a valid path.

tests/unit/test_download.py
```diff
@@ -65,6 +69,58 @@ def read(self, *args, **kwargs):
...
+        self.filename = os.path.join(self.tempdir, 'myfile')
+
+        self.call_args = CallArgs(fileobj=self.filename)
+        self.future = self.get_transfer_future(self.call_args)
+
+    def tearDown(self):
+        super(TestDownloadFilenameOutputManager, self).tearDown()
+        shutil.rmtree(self.tempdir)
+
+    def test_is_compatible(self):
```
JordonPhillips: Could we also test for something that's incompatible, like a `StringIO` or something.

s3transfer/download.py
```diff
@@ -131,6 +131,27 @@ def _get_temp_fileobj(self):
         return f
 
 
+class DownloadSeekableOutputManager(DownloadOutputManager):
+    @classmethod
+    def is_compatible(self, download_target):
+        return (
+            hasattr(download_target, 'seek') and
+            hasattr(download_target, 'tell')
```
JordonPhillips: An `io.IOBase` will have `seek`, but is not necessarily be seek-able or tell-able. You need to call `io.IOBase.seekable` to be able to tell.

tests/functional/test_download.py
```diff
@@ -145,6 +145,19 @@ def test_download_temporary_file_does_not_exist(self):
         possible_matches = glob.glob('%s*' % self.filename + os.extsep)
         self.assertEqual(possible_matches, [])
 
+    def test_download_for_fileobj(self):
```
JordonPhillips: Could you add a test for a non-file file-like object, such as `StringIO` or `BytesIO`?

tests/unit/test_download.py
```diff
@@ -121,6 +128,40 @@ def test_get_final_io_task(self):
...
+        self.fileobj = open(self.filename, 'wb')
+
+        self.call_args = CallArgs(fileobj=self.fileobj)
+        self.future = self.get_transfer_future(self.call_args)
+
+    def tearDown(self):
+        self.fileobj.close()
+        super(TestDownloadSeekableOutputManager, self).tearDown()
+
+    def test_is_compatible(self):
```
JordonPhillips: Could we get some tests for incompatible types, for instance a non-seekable `io.IOBase`

s3transfer/download.py
```diff
@@ -38,9 +39,145 @@
...
+    def get_fileobj_for_io_writes(self, transfer_future):
+        # Return the fileobj provided to the future.
+        return transfer_future.meta.call_args.fileobj
+
+    def get_final_io_task(self):
+        # This task will serve the purpose of signaling when all of the io
+        # writes have finished so done callbacks can be called.
+        return CompleteDownloadNOOPTask(
+            transfer_coordinator=self._transfer_coordinator,
+            is_final=True
```
jamesls: Is it ever valid to create a `CompleteDownloadNOOPTask` with `is_final=False`.  Should is_final be built into the `CompleteDownloadNOOPTask`.

s3transfer/download.py
```diff
@@ -38,9 +39,145 @@
...
+
+        :type transfer_future: s3transfer.futures.TransferFuture
+        :param transfer_future: The future associated with upload request
+
+        returns: A file-like object to write to
+        """
+        raise NotImplementedError('must implement get_fileobj_for_io_writes()')
+
+    def get_final_io_task(self):
+        """Get the final io task to complete the download
```
jamesls: It would be helpful to document why this is required for subclasses to implement.  The doc string doesn't give a whole lot of information, and I can see myself forgetting why we have to implement this for seekable fileobjs that just return a noop task.

s3transfer/download.py
```diff
@@ -38,9 +39,145 @@
...
+            transfer_coordinator=self._transfer_coordinator,
+            is_final=True
+        )
+
+
 class DownloadSubmissionTask(SubmissionTask):
     """Task for submitting tasks to execute a download"""
 
+    def _get_download_output_manager_cls(self, transfer_future):
+        """Retieves a class for managing output for a download
```
jamesls: s/Retieves/Retrieves/

jamesls:
Pretty straightfoward, no major issues for me.  I agree with @JordonPhillips's feedback about seekable and test improvements.  Otherwise :shipit:

s3transfer/download.py
kyleknap: seekable() is not available on the fileobj returned from ``open()`` in Python 2. How about I just check that tell() does not return an IOError?

JordonPhillips: Could you make it a compat function? I feel like we should use it if it's available.

tests/unit/test_download.py
kyleknap: Those would actually be compatible as it is seekable. I will create classes that are not seekable.

s3transfer/download.py
kyleknap: Yeah that fine. I would just need to figure out how ``seekable()`` is implemented in python. I think it is written in C so it may be interesting.

jamesls: Looks like it just tries lseek (seek). https://github.com/python/cpython/blob/master/Modules/_io/fileio.c#L609-L625

kyleknap: Nice. I will do that then.

s3transfer/download.py
kyleknap: I can do that.

kyleknap:
@jamesls @JordonPhillips 
I updated the following based on feedback:
1) Created a ``seekable()`` function in ``compat.py`` and ported uploads and downloads to use it.
2) Updated docs for ``get_final_io_task`` and made the ``CompleteDownloadNOOPTask`` ``is_final=True`` by default.

Let me know what you think.

s3transfer/compat.py
```diff
@@ -50,3 +50,27 @@ def accepts_kwargs(func):
...
+
+    :param fileobj: The file-like object to determine if seekable
+
+    :returns: True, if seekable. False, otherwise.
+    """
+    try:
+        # If the fileobj has a seekable attr, try calling the seekable()
+        # method on it.
+        if hasattr(fileobj, 'seekable'):
+            return fileobj.seekable()
```
jamesls: Can seekable actually raise an IOError/OSError?  Does it need to be under the try?

kyleknap: It does not need to be. I can move it out. It should always return True or False

jamesls:
:shipit:, had a question, but otherwise looks good to me.

jamesls:
:shipit:

JordonPhillips:
```

                       _____|\
                  _.--| SSt |:
                 <____|.----||
                        .---''---,
                         ;..__..'    _...
                       ,'/  ;|/..--''    \
                     ,'_/.-/':            :
                _..-'''/  /  |  \    \   _|/|
               \      /-./_ \;   \    \,;'   \
               ,\    / \:  `:\    \   //    `:`.
             ,'  \  /-._;   | :    : ::    ,.   .
           ,'     ::   /`-._| |    | || ' :  `.`.)
        _,'       |;._:: |  | |    | `|   :    `'
      ,'   `.     /   |`-:_ ; |    |  |  : \
      `--.   )   /|-._:    :          |   \ \
         /  /   :_|   ;`-._;   __..--';    : :
        /  (    ;|;-./_  _/.-:'o |   /     ' |
       /  , \._/_/_./--''/_|:|___|_,'        |
      :  /   `'-'--'----'---------'          |
      | :     O ._O   O_. O ._O   O_.      ; ;
      : `.      //    //    //    //     ,' /
    ~~~`.______//____//____//____//_______,'~
              //    //~   //    //
       ~~   _//   _//   _// ~ _//     ~
     ~     / /   / /   / /   / /  ~      ~~
          ~~~   ~~~   ~~~   ~~~
```

PR #19, opened 2016-06-29
Add TransferManagerController for Cntrl-C

kyleknap:
The abstraction serves the purpose of keeping track all in progress transfers such that:

   * We can Cntrl-C for TransferFuture.result().

   * Introduce a context handler on the TransferManager that wait for all
     TransferFutures to complete, can be quickly cancel all TransferFutures
     if encounters Cntrl-C, and can cancel all inprogress TransferFutures if
     an exception is raised in the middle of the with statement as if there
     is an exception the goal should be to exit as quickly and cleanly as
     possible while still raising the exception hit.

Note that Cntrl-C for uploads are still pretty slow. We will need to wrap the body that is sent with something that will interrupt the sending of the body.

@jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -14,6 +14,7 @@ from concurrent import futures
 from collections import namedtuple
 import copy
 import logging
+import sys
 import threading
 
 from s3transfer.utils import FunctionContainer
@@ -177,22 +178,18 @@ class TransferCoordinator(object):
                 self._exception = exception
                 self._status = 'failed'
 
-    def result(self, sleep_intervals=0.2):
+    def result(self):
         """Waits until TransferFuture is done and returns the result
 
         If the TransferFuture succeeded, it will return the result. If the
         TransferFuture failed, it will raise the exception associated to the
         failure.
-
-        :type sleep_intervals: int or float
-        :param sleep_intervals: The amount of time to sleep each interval
-            while waiting for the transfer to complete.
         """
-        # We add an interval to sleep because we want to allow the wait
-        # to be interrupted just in case a KeyboardInterrupt happened.
-        # Doing a wait() with no time to sleep cannot be interrupted.
-        while not self._done_event.is_set():
-            self._done_event.wait(sleep_intervals)
+        # Doing a wait() with no timeout cannot be interrupted in python2 but
+        # can be interrupted in python3 so we just wait with the largest
+        # possible value integer value, which is on the scale of billions of
+        # years...
+        self._done_event.wait(sys.maxint)
 
         # Once done waiting, raise an exception if present or return the
         # final result.
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -14,9 +14,9 @@ from concurrent import futures
 from collections import namedtuple
 import copy
 import logging
-import sys
 import threading
 
+from s3transfer.compat import MAXINT
 from s3transfer.utils import FunctionContainer
 
 
@@ -189,7 +189,7 @@ class TransferCoordinator(object):
         # can be interrupted in python3 so we just wait with the largest
         # possible value integer value, which is on the scale of billions of
         # years...
-        self._done_event.wait(sys.maxint)
+        self._done_event.wait(MAXINT)
 
         # Once done waiting, raise an exception if present or return the
         # final result.
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -28,7 +28,7 @@ class TransferFuture(object):
 
         :type meta: TransferMeta
         :param meta: The metadata associated to the request. This object
-            is visible to the requester..
+            is visible to the requester.
 
         :type coordinator: TransferCoordinator
         :param coordinator: The coordinator associated to the request. This
@@ -61,7 +61,14 @@ class TransferFuture(object):
         TransferFuture failed, it will raise the exception associated to the
         failure.
         """
-        return self._coordinator.result()
+        try:
+            # Usually the result() method blocks until the transfer is done,
+            # however if a KeyboardInterrupt is raised we want want to exit
+            # out of this and propogate the exception.
+            return self._coordinator.result()
+        except KeyboardInterrupt as e:
+            self.cancel()
+            raise e
 
     def cancel(self):
         """Cancels the request associated with the TransferFuture"""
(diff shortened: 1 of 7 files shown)


s3transfer/futures.py
```diff
@@ -170,14 +177,25 @@ def set_exception(self, exception):
                 self._exception = exception
                 self._status = 'failed'
 
-    def result(self):
+    def result(self, sleep_intervals=0.2):
```
kyleknap: Not a big fan of having to do sleep intervals. Anyone have good ideas of how we can better handle this and if not what a more appropriate time should be? The reason I have to do this is a straight Event.wait() will consume the KeyboardInterrupt and will not raise it till the event gets set to done, which would be the entirety of the transfer...

s3transfer/manager.py
```diff
@@ -378,11 +389,106 @@ def _register_handlers(self):
...
+        """Cancels all inprogress transfers
+
+        This cancels the inprogress transfers by calling cancel() on all
+        tracked transfer coordinators.
+        """
+        with self._lock:
+            for transfer_coordinator in self._active_transfer_coordinators:
+                transfer_coordinator.cancel()
+
+    def wait(self, sleep_intervals=0.2):
```
kyleknap: Same thing here. I am up for suggestions for how to avoid sleeping. I am not sure if using an event can be used as well for the reason that it consumes KeyboardInterrupts and fails to raise it till all tasks are done.

kislyuk: Have you tried to pass a timeout argument to wait? You can pass sys.maxint or similar to effectively wait forever. In Python 2.7, waiting on a thread without a timeout causes an uninterruptible wait. In Python 3, the wait is interruptible. See https://github.com/agronholm/pythonfutures/issues/25.

kyleknap: Ahh I did not realize just the passing in any timeout enables the ability to interrupt in python2. I sort of had that in the TransferCoordinator.result(): https://github.com/boto/s3transfer/pull/19/files/3a9b618d4eeeb5f2ede959e5b43bbfcc78cb7a28#diff-740cbe5420ec5bbd23092951593c146dR195, but I made it more of a pulling mechanism, which is not great as it is less efficient. Thanks for the suggestion! Very helpful.

kislyuk: Yeah, this is something it took me a long time to figure out myself in a similar application, since the behavior doesn't seem to be documented anywhere.

I'm hoping this will become my go-to library for interfacing with S3 for bulk data transfers. Good to see it coming together. Please consider tracking performance and memory use in the test suite too, to avoid regressions. I've seen some severe transport performance regressions make it into releases in even the highest profile projects like requests and node.

kyleknap: Yeah that was definitely not obvious.

And yep. Along with functionality, performance is one of the items that we want to make sure we really nail as we want to use this in boto3 and the AWS CLI.

I really like the idea of having performance tests. I think it is something that we should add once the framework is built up and are happy with the performance.

s3transfer/manager.py
```diff
@@ -378,11 +389,114 @@ def _register_handlers(self):
...
+            # the inflight transfers before shutting down to speed that
+            # process up.
+            self._canceler.cancel()
+            # Shutdown all of the executors.
+            self._submission_executor.shutdown()
+            self._request_executor.shutdown()
+            self._io_executor.shutdown()
+
+
+class TransferCoordinatorCanceler(object):
```
jamesls: This is more like a coordinator-coordinator (don't call it that).  But it seems to provide functionality that can track multiple coordinators.  In addition to cancelling you can also wait on coordinators as well.  I think the name should reflect that.

s3transfer/manager.py
```diff
@@ -378,11 +389,114 @@ def _register_handlers(self):
...
+            # If Keyboard interrupt is raised while waiting for
+            # the result, then exit out of the wait and raise the
+            # exception
+            raise
+        except Exception:
+            # A general exception could have been thrown because
+            # of result(). We just want to ignore this and continue
+            # because we at least know that the transfer coordinator
+            # has completed.
+            pass
```
jamesls: Should we not be raising some sort of exception though at the end to indicate the errors occurred?

s3transfer/manager.py
kyleknap: Yeah... I was having troubles coming up with a name. Any ideas? I originally was going to call it a TransferManagerController but that is vague.

s3transfer/manager.py
kyleknap: That would mean that any result would be propogated in the the context handler. So if I have logic like this:
```py
with manager as m:
     future = m.upload('file', 'fakebucket', 'key')

future.result()
```
The error will actually happen exiting the context manager instead of at the ``future.result()``. This is a larger issue if you are trying to batches of transfers and then want to analyze all the results at the end.

The only reason I am using ``result()`` is because it is the only method on the transfer coordinator that is hooked up to the done event. If you are fine with it, I can add a new ``wait_until_done()`` method on the ``TransferCoordinator`` if that makes sense?

s3transfer/manager.py
kyleknap: Here are other names I can think of:

* TransferCoordinatorManager
* TransferCoordinatorStatusManager
* TransferCoordinatorTracker

Any ideas on these suggestions? I can am leaning toward TransferCoordinatorManager. The other permutation is a bit specific and this class could possibly grow as it can access all public methods of the transfer coordinator.

s3transfer/manager.py
jamesls: I guess it depends on what we consider is the contract of using the transfer manager as a context manager.

Is your code sample above even something we should support?  Seems like if you're using a context manager, you should ask for the future results within the context.  In which case I would expect an exception to be thrown exiting the with statement if there was no `.result()` call within the `with` statement.

s3transfer/manager.py
jamesls: Controller seemed reasonable.  But I think that's just my bias that I dislike any class name with `Manager` in it, as I find that name to be vague.

kyleknap: Cool I will call it ``TransferCoordinatorController`` then

s3transfer/manager.py
kyleknap: @jamesls We talked about this where I think this should be consistent with the current interface of executors in python:
```py
In [1]: from concurrent.futures import ThreadPoolExecutor

In [2]: def raise_exception():
   ...:     raise Exception()
   ...: 

In [3]: with ThreadPoolExecutor(1) as executor:
   ...:     future = executor.submit(raise_exception)
   ...:     

In [4]: future.result()
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-4-a3ab6647db03> in <module>()
----> 1 future.result()
```
I do agree though that we should have some interface for enabling fast exits of the transfer manager if any of the transfers fail such as:
```py
with TransferManager(client) as m:
   with m.raise_if_failures():
       m.upload('myfile', 'mynonexistentbucket', 'key') 
```
Probably should come up with a better name but this would be nice because you can be specific on what transfers you want a hard failure for.

Let me know what you think.

kyleknap:
@jamesls Should be good to look at again. I updated the name of the class and expanded on the context handler comments on if we should fail fast on future.result() in the context handler.

s3transfer/manager.py
jamesls: Seems reasonable.  Let's make sure we document/comment that in the code.

jamesls:
:shipit:

kyleknap:
Cool. Updated the documentation. Merging.

PR #21, opened 2016-07-01
Wrap upload bodies with InterruptReader

kyleknap:
This is the second part needed to quickly cancel all transfers.

This enables the interruption of a socket while it reads so that the entire body does not have to be uploaded before a transfer can be cancelled. So you can do something like this:
```py
import boto3

client = boto3.client('s3')

with TransferManager(client) as m:
    future1 = m.upload('100MB', 'mybucket', '100MB-upload')
    future1.result()
```
And should be able to get it cancel using Cntrl-C within a couple of seconds.

I also considered wrapping the fileobj ``PutObjectTask`` and ``UploadPartTask`` but that would require me to implement a lot of the interfaces provided by ``ReadFileChunk``, which is the current top wrapper layer, and that did not make sense because I would have to enable and disable callbacks.

I am not sure if it is strange that I pass the TransferCoordinator now to the UploadInputManagers, but I figured it is acceptable because it now matches the initializer for the DownloadOutputManager.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/upload.py
+++ b/s3transfer/upload.py
@@ -24,6 +24,49 @@ from s3transfer.utils import get_callbacks
 from s3transfer.utils import DeferredOpenFile
 
 
+class InterruptReader(object):
+    """Wrapper that can interrupt reading using an error
+
+    It uses a transfer coordinator to propogate an error if it notices
+    that a read is being made while the file is being read from.
+
+    :type fileobj: file-like obj
+    :param fileobj: The file-like object to read from
+
+    :type transfer_coordinator: s3transfer.futures.TransferCoordinator
+    :param transfer_coordinator: The transfer coordinator to use if the
+        reader needs to be interrupted.
+    """
+    def __init__(self, fileobj, transfer_coordinator):
+        self._fileobj = fileobj
+        self._transfer_coordinator = transfer_coordinator
+
+    def read(self, amount=None):
+        # If there is an exception, then raise the exception.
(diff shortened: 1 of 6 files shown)


s3transfer/upload.py
```diff
@@ -24,6 +24,49 @@
...
+        self._transfer_coordinator = transfer_coordinator
+
+    def read(self, amount=None):
+        # If there is an exception, then raise the exception.
+        # We raise an error instead of returning no bytes because for
+        # requests where the content length and md5 was sent, it will
+        # cause md5 mismatches and retries as there was no indication that
+        # the stream being read from encountered any issues.
+        if self._transfer_coordinator.exception:
+            raise self._transfer_coordinator.exception
```
jamesls: This is going to acquire a shared lock twice.  I would imagine this could slow things down as you ramp up your parallelism.

I don't think `self._exception` needs to be protected by a lock does it?  Reading from it should be thread safe.  What do you think about removing the lock?

jamesls:
Looks good, just had a question about a shared lock.

s3transfer/upload.py
kyleknap: Now that I look at it. It does not need a lock. I am just setting/getting a value. That should be thread safe. I will remove it.

kyleknap:
@jamesls I updated the code. You were right. The lock was not needed.

jamesls:
:shipit:

PR #22, opened 2016-07-06
Fix issue with tests and added more logging

kyleknap:
The issue was that the KeyboardInterrupt would be called during the
execution of the CompleteMultipartUpload and as a result the
AbortMultipartUpload would not delete the object because the multipart
upload already completed. The solution was to make the sleep time shorter to
help better ensure the abort happens before the complete multipart upload
and log when KeyboardInterrupts are sent and when coordinators are cancelled.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -206,6 +206,7 @@ class TransferCoordinator(object):
         """Cancels the TransferFuture"""
         if not self.done():
             with self._lock:
+                logger.debug('TransferCoordinator cancel() called')
                 self._exception = futures.CancelledError
                 self._status = 'cancelled'
 
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -165,12 +165,17 @@ class TransferCoordinator(object):
     def set_result(self, result):
         """Set a result for the TransferFuture
 
-        Implies that the TransferFuture succeeded.
+        Implies that the TransferFuture succeeded. This will always set a
+        result because it is invoked on the final task where there is only
+        ever one final task and it is ran at the very end of a transfer
+        process. So if a result is being set for this final task, the transfer
+        succeeded even if something came a long and canceled the transfer
+        on the final task.
         """
-        if not self.done():
-            with self._lock:
-                self._result = result
-                self._status = 'success'
+        with self._lock:
+            self._exception = None
+            self._result = result
+            self._status = 'success'
 
     def set_exception(self, exception):
         """Set an exception for the TransferFuture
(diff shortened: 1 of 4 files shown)


kyleknap:
Alright I updated the code such that ``set_result`` clobbers any exception. The reason that we do this now is that the ``set_result`` is only ever ran on the final task and if the final task succeeds. If that is the case, the entire transfer succeeded because the final task is only ran if its dependent tasks passed and it is always ran by itself at the end. As a result, a ``cancel()`` should not be reflected in ``future.result()`` if the cancel did not actually cancel the transfer and that final task succeeded.

jamesls:
:shipit:

Should we have the same behavior with `set_exception()`.  What happens if the final task raises an exception but something has already cancelled the transfer (e.g. a Ctrl-C)?

kyleknap:
I do not think we should have the same behavior for ``set_exception``. For the following reasons:

1) ``'cancelled'`` and ``'failed'`` are essentially the same state as they both will throw an exception on ``future.result()`` and run any cleanup methods because of it. The only difference between failed and cancelled is that it gives you a little more grainularity on if it was the user that explicitly caused the task to fail or not. Thus knowing the difference between 'cancelled' and 'failed' is not as important as the difference between 'cancelled' and 'success'
2) Note that the ``set_exception`` can be ran at any time for any task. So it may be misleading on what caused the task to not be successful. If ``set_exception`` can override cancels and a cancel initiates the shutdown of a task, that would propogate the exception instead of the cancelled exception (even though the cancel happened first). Ideally, when the state of the future is not success it should always raise what was the incident to kick it off and by not allowing ``set_exception`` to override ``cancel``, it is better able to do that.

No matter in the end, I do not think we can perfectly determine if a cancel error or another error should be the one propogated if they nearly happened at the same time, but in the end, it is not really a big issue because failed and cancelled state have essentially the same behavior and workflow.

@jamesls does that make sense?

jamesls:
First, I don't think this discussion should block getting this merged.  The fix as is, is good and we can address the `set_exception`  behavior in a separate thread.  But I still think the existing behavior is confusing, assuming I'm not missing anything.

From what I understand:

* A thread comes along and cancels the task.
* Whether or not that cancellation "sticks" depends on what happens in the final task.
* If the final task succeeds, then `set_result()` overrides and the cancel for that task is essentially ignored.
* If the final task fails (say API error or something), then the cancellation sticks and we ignore the error raised in the final task.

That still seems like a weird consistency to me.

kyleknap:
Ah I see what you mean. The unfortunate part about this is that making that change would not be as easy as ``set_result``. We would have to add an override flag to ``set_exception`` to make that work and only set the flag on the final task.

PR #23, opened 2016-07-06
Add support for non-seekable streaming uploads

JordonPhillips:
The tricky part of this change is that existing code relies on the
ability to get the length of the object being uploaded, which is
not possible for non-seekable objects. To work around that issue,
this will read data into memory up to the configured multipart
threshold to determine if it requires a multipart upload. That data
will then be used to generate the initial parts.

cc @kyleknap @jamesls

### DIFF
--- a/s3transfer/upload.py
+++ b/s3transfer/upload.py
@@ -14,7 +14,7 @@ import math
 
 from botocore.compat import six
 
-from s3transfer.compat import seekable
+from s3transfer.compat import seekable, readable
 from s3transfer.futures import IN_MEMORY_UPLOAD_TAG
 from s3transfer.tasks import Task
 from s3transfer.tasks import SubmissionTask
@@ -82,6 +82,10 @@ class UploadInputManager(object):
     that may be accepted. All implementations must subclass and override
     public methods from this class.
     """
+    def __init__(self, osutil, transfer_coordinator):
+        self._osutil = osutil
+        self._transfer_coordinator = transfer_coordinator
+
     @classmethod
     def is_compatible(cls, upload_source):
         """Determines if the source for the upload is compatible with manager
@@ -163,13 +167,12 @@ class UploadInputManager(object):
         """
         raise NotImplementedError('must implement yield_upload_part_bodies()')
 
+    def _wrap_with_interrupt_reader(self, fileobj):
+        return InterruptReader(fileobj, self._transfer_coordinator)
+
 
 class UploadFilenameInputManager(UploadInputManager):
     """Upload utility for filenames"""
-    def __init__(self, osutil, transfer_coordinator):
-        self._osutil = osutil
-        self._transfer_coordinator = transfer_coordinator
-
     @classmethod
     def is_compatible(cls, upload_source):
         return isinstance(upload_source, six.string_types)
@@ -227,9 +230,6 @@ class UploadFilenameInputManager(UploadInputManager):
                 full_file_size=full_size, callbacks=callbacks)
             yield part_number, read_file_chunk
 
-    def _wrap_with_interrupt_reader(self, fileobj):
-        return InterruptReader(fileobj, self._transfer_coordinator)
-
     def _get_deferred_open_file(self, fileobj, start_byte):
         fileobj = DeferredOpenFile(fileobj, start_byte)
         fileobj.OPEN_METHOD = self._osutil.open
@@ -249,10 +249,10 @@ class UploadFilenameInputManager(UploadInputManager):
 
 
 class UploadSeekableInputManager(UploadFilenameInputManager):
-    """Upload utility for am open file object"""
+    """Upload utility for an open file object"""
     @classmethod
     def is_compatible(cls, upload_source):
-        return seekable(upload_source)
+        return readable(upload_source) and seekable(upload_source)
 
     def stores_body_in_memory(self, operation_name):
         if operation_name == 'put_object':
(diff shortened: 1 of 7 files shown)


tests/functional/test_upload.py
```diff
@@ -296,6 +297,17 @@ def test_upload_for_seekable_filelike_obj(self):
         self.assert_expected_client_calls_were_correct()
         self.assert_upload_part_bodies_were_correct()
 
+    def test_upload_for_non_seekable_filelike_obj(self):
```
kyleknap: We probably will want to add one for the non mulitpart upload case as well.

s3transfer/upload.py
```diff
@@ -286,6 +286,157 @@ def _get_put_object_fileobj(self, fileobj):
...
+class UploadNonSeekableInputManager(UploadInputManager):
+    """Upload utility for a file-like object that cannot seek."""
+    def __init__(self, osutil, transfer_coordinator):
+        super(UploadNonSeekableInputManager, self).__init__(
+            osutil, transfer_coordinator)
+        self._initial_data = None
+
+    @classmethod
+    def is_compatible(cls, upload_source):
+        return not seekable(upload_source)
```
kyleknap: So I am not sure if the qualifier here should be weather it is seekable or not. Technically, you could put a seekable file through this Manager and it should work. I think the compatibility here should be if the file is readable as if it fails that the ``UploadSubmissionTask`` will throw an error as this is the last one in the input manager resolver chain.

s3transfer/upload.py
```diff
@@ -286,6 +286,157 @@ def _get_put_object_fileobj(self, fileobj):
...
+        # No-op because there is no way to do this short of reading the entire
+        # body into memory.
+        return
+
+    def requires_multipart_upload(self, transfer_future, config):
+        # This is tricky to determine in this case because we can't know how
+        # large the input is. So to figure it out, we read data into memory
+        # up until the threshold and compare how much data was actually read
+        # against the threshold.
+        fileobj = transfer_future.meta.call_args.fileobj
```
kyleknap: If the user provided a size themselves, I would say we do not stream the file in. So I would make sure you have this check at beginning: https://github.com/boto/s3transfer/blob/develop/s3transfer/upload.py#L186 and then do the streaming.

s3transfer/upload.py
```diff
@@ -286,6 +286,157 @@ def _get_put_object_fileobj(self, fileobj):
...
+        callbacks = get_callbacks(transfer_future, 'progress')
+        part_number = 0
+
+        # First go over the parts generated by calling
+        # `requires_multipart_upload` if any.
+        if self._initial_data is not None:
+            parts = self._yield_first_parts(transfer_future, config, callbacks)
+            for part in parts:
+                part_number += 1
+                yield part_number, part
```
kyleknap: We talked about this but it would be great if we had all of this logic under one loop and one yield from. The two seperate loops and the three places to yield from is a bit complicated.

s3transfer/upload.py
```diff
@@ -286,6 +286,157 @@ def _get_put_object_fileobj(self, fileobj):
...
+        :type config: s3transfer.manager.TransferConfig
+        :param config: The config associated to the transfer manager.
+
+        :type callbacks: list
+        :param callbacks: The callbacks associated with the transfer future.
+
+        :return: Generator which generates part bodies from the initial data.
+        """
+        try:
+            fileobj = transfer_future.meta.call_args.fileobj
```
kyleknap: These three lines do not need to be in the try except block.

s3transfer/upload.py
```diff
@@ -286,6 +286,157 @@ def _get_put_object_fileobj(self, fileobj):
...
+                    yield self._wrap_part_data(
+                        self._initial_data[i:part_bound], callbacks)
+        except GeneratorExit as e:
+            # Zero out the initial data now that we're done generating parts
+            # from it so that we don't hold onto those bytes any longer
+            # than we need to.
+            self._initial_data = None
+            raise e
+
+    def _wrap_part_data(self, data, callbacks, size=None):
```
kyleknap: I think we can call it something more general like ``_wrap_data`` as this is used for PutObject and UploadPart.

tests/integration/test_upload.py
```diff
@@ -26,22 +27,23 @@ def setUp(self):
         self.config = TransferConfig(
             multipart_threshold=self.multipart_threshold)
 
+    def create_file(self, size, name=''):
```
kyleknap: Let's call it ``get_input_fileobj()`` or something along those lines. ``create_file`` for me seems like you are intending to create a file each time and in reality for the file-like objects you are just creating a file-like object without writing to disk.

tests/unit/test_upload.py
```diff
@@ -262,6 +266,71 @@ def test_stores_bodies_in_memory_upload_part(self):
...
+
+        parts = list(self.upload_input_manager.yield_upload_part_bodies(
+            self.future, self.config))
+
+        size = self.config.multipart_chunksize
+        num_parts = len(self.content) / size
+        num_parts = math.ceil(num_parts)
+        self.assertEqual(len(parts), num_parts)
+
+        for i, part in enumerate(parts[:-1]):
```
kyleknap: I can follow this but it took a little bit of time. Comments would have helped especially in the for loop, the initial part calculations, and the final part assertion.

kyleknap:
Implementation looks good. There was a little bit of refactoring to simplify the generator that I pointed out that we can do and a couple edge cases that looks like you may have missed, but looks good in general.

JordonPhillips:
Updated

kyleknap:
Looks like python3 tests are failing...

JordonPhillips:
Fixed

kyleknap:
Nice. Really like the refactoring you did. Makes it much easier to follow. :ship:

kyleknap:
Also before merging, it would not hurt to do a performance test to see how the speed compares to the seekable and filename case.

JordonPhillips:
Rebased and squashed commits

JordonPhillips:
Performance differences seem to be negligible.

Time in seconds for a single 7 byte upload:

| Actual File                   | Seekable File-Like         | Non-Seekable File-Like |
| ----------------------------- | ------------------------------- | ------------------------------- |
| 0.1925144976199954 | 0.22886600455996814 | 0.15985330105002504 |

Time in seconds for a multipart upload of 5MB + 1 byte with a 5MB chunksize and threshold:

| Actual File                   | Seekable File-Like         | Non-Seekable File-Like |
| ----------------------------- | ------------------------------- | ------------------------------- |
| 0.8002007092098938 | 0.7937404662198969 | 0.8091048456198768 |

kyleknap:
Thanks. :ship:

PR #24, opened 2016-07-06
Allow users to inject OSUtils

kyleknap:
This is mainly needed for the boto3 port as it allows you to pass your own ``OSUtils`` object. In injecting it, I also noticed that the uploads were not actually using the ``OSUtils.open`` so I fixed that as well.

cc @jamesls @JordonPhillips


jamesls:
:shipit:

PR #25, opened 2016-07-07
Add support for non-seekable downloads

jamesls:
This adds support for non seekable downloads.  It's not 100% ready to go, but I'd like to get some design feedback before I get too far along.  It's also still missing an integration test.

I'd recommend reviewing the commits in order, I tried to break them down logically.

The big change is that I moved the IO task submission into the download manager.  Doing so let me put all the non seekable logic in a new download output manager rather than cluttering the GetObjectTask.  This also means that the download output managers take an io executor.

I also had to add an abstraction to buffer/defer IO writes until we could queue them in sequential order.

The other missing piece is that there's nothing here that's capping the max memory allowed for buffering IO writes.  That will likely be in another pull request.

To test non-seekable, I just used:

```
class NonSeekable(object):
    def __init__(self, f):
        self.f = f

    def read(self, amt=None):
        return self.f.read(amt)

    def write(self, content):
        self.f.write(content)

    def close(self):
        self.f.close()
```

Perf is about the same as a seekable fileobj:

```
From seekable fileobj:
Queued download time            : 0.0012
Download complete time          : 22.9263

STREAMING

Queued download time            : 0.0012
Download complete time          : 22.6799

```

cc @kyleknap @JordonPhillips

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -14,6 +14,9 @@ import logging
 import os
 import socket
 import math
+import threading
+import heapq
+
 
 from botocore.compat import six
 from botocore.exceptions import IncompleteReadError
@@ -176,6 +179,54 @@ class DownloadSeekableOutputManager(DownloadOutputManager):
 
 
 
+class DownloadNonSeekableOutputManager(DownloadOutputManager):
+    def __init__(self, osutil, transfer_coordinator, io_executor,
+                 defer_queue=None):
+        super(DownloadNonSeekableOutputManager, self).__init__(
+            osutil, transfer_coordinator, io_executor)
+        if defer_queue is None:
+            defer_queue = DeferQueue()
+        self._defer_queue = defer_queue
+        self._io_submit_lock = threading.Lock()
+
+    @classmethod
+    def is_compatible(cls, download_target):
+        # We're lying a bit here.  We actually *are* compatible
+        # with seekable fileobjs, we'll still queue the IO writes
+        # in order.  Technically works, but this is not ideal behavior when you
+        # could just use DownloadSeekableOutputManager.
+        # Perhaps this method makes more sense as "should_use()"
+        return (
+            not seekable(download_target) and
+            hasattr(download_target, 'write')
+        )
+
+    def get_fileobj_for_io_writes(self, transfer_future):
+        return transfer_future.meta.call_args.fileobj
+
+    def get_final_io_task(self):
+        return CompleteDownloadNOOPTask(
+            transfer_coordinator=self._transfer_coordinator)
+
+    def queue_file_io_task(self, fileobj, data, offset):
+        with self._io_submit_lock:
+            writes = self._defer_queue.request_writes(offset, data)
+            for write in writes:
+                data = write['data']
+                logger.debug("Queueing IO offset %s for fileobj: %s",
+                            write['offset'], fileobj)
+                future = self._io_executor.submit(
+                    IOStreamingWriteTask(
+                        self._transfer_coordinator,
+                        main_kwargs={
+                            'fileobj': fileobj,
+                            'data': data,
+                        }
+                    )
+                )
+                self._transfer_coordinator.add_associated_future(future)
+
+
 class DownloadSubmissionTask(SubmissionTask):
     """Task for submitting tasks to execute a download"""
 
@@ -192,6 +243,7 @@ class DownloadSubmissionTask(SubmissionTask):
         download_manager_resolver_chain = [
             DownloadFilenameOutputManager,
             DownloadSeekableOutputManager,
+            DownloadNonSeekableOutputManager,
         ]
 
         fileobj = transfer_future.meta.call_args.fileobj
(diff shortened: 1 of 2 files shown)


s3transfer/download.py
```diff
@@ -79,6 +83,27 @@ def get_fileobj_for_io_writes(self, transfer_future):
...
 
+    def queue_file_io_task(self, fileobj, data, offset):
+        """Queue IO write for submission to the IO executor.
+
+        This method accepts an IO executor and information about the
+        downloaded data, and handles submitting this to the IO executor.
+
+        This method may defer submission to the IO executor if necessary.
+
+        """
```
kyleknap: I agree with this approach as we really would not want to pass in a reference to the ``GetObjectTask`` to invoke the ``_submit_task``. But should we put in a log statement as well to achieve better parity with the ``_submit_task``: https://github.com/boto/s3transfer/blob/develop/s3transfer/tasks.py#L200?

s3transfer/download.py
```diff
@@ -153,6 +178,55 @@ def get_final_io_task(self):
...
+            defer_queue = DeferQueue()
+        self._defer_queue = defer_queue
+        self._io_submit_lock = threading.Lock()
+
+    @classmethod
+    def is_compatible(cls, download_target):
+        # We're lying a bit here.  We actually *are* compatible
+        # with seekable fileobjs, we'll still queue the IO writes
+        # in order.  Technically works, but this is not ideal behavior when you
+        # could just use DownloadSeekableOutputManager.
```
kyleknap: Yeah so the way I consider ``is_compatible`` is simply if the manager could work with the fileobj given (it does not necessarily mean it is efficient). And in the end it is up to the SubmissionTask to determine, based on its own order of preference and what managers are compatible with the fileobj, what manager to use. So for your logic you would probably want to check if it is writeable. Jordon is doing a similar thing in the nonseekable case.

s3transfer/download.py
```diff
@@ -153,6 +178,55 @@ def get_final_io_task(self):
...
+        # Perhaps this method makes more sense as "should_use()"
+        return (
+            not seekable(download_target) and
+            hasattr(download_target, 'write')
```
kyleknap: Should we add a writeable() back compat that will check:
```py
if hasattr(download_target, 'writeable'):
    return download_target.writeable()
return hasattr(download_target, 'write)
```
Jordon is doing the same thing for nonseekable uploads but for ``readable``

s3transfer/download.py
```diff
@@ -475,3 +558,38 @@ def __init__(self, transfer_coordinator, main_kwargs=None,
...
+        method will return all applicable writes.
+
+        This is done with 1 method call so you don't have to
+        make two method calls (put(), get()) which acquires a lock
+        each method call.
+
+        """
+        writes = []
+        heapq.heappush(self._writes, (offset, data))
+        while self._writes and self._writes[0][0] == self._next_offset:
```
kyleknap: Can you explain how retries are handled? For example, we start adding data to the heap and that bumps up the ``self._next_offset``. Then let's say that the download got retried and we start at the beginning again with an offset less than the ``self._next_offset``. It looks like the queue will never be cleared because the reset offset will be smaller than the expected offse and the while loop will never be entered. To fix this, I think you need to check if the given offset is less than the _next_offset and discard it if so.

tests/unit/test_download.py
```diff
@@ -169,6 +174,52 @@ def test_get_final_io_task(self):
...
+                pass
+
+        f = NonSeekable()
+        self.assertTrue(self.download_output_manager.is_compatible(f))
+
+    def test_no_compatible_with_bytesio(self):
+        self.assertFalse(
+            self.download_output_manager.is_compatible(six.BytesIO()))
+
+    def test_submit_writes_from_internal_queue(self):
```
kyleknap: Should we have a test for ``queue_io_file_task`` for the other managers?

tests/unit/test_download.py
```diff
@@ -350,6 +405,31 @@ def tearDown(self):
         self.files.remove_all()
 
 
+class TestIORenameFileTask(BaseIOTaskTest):
```
kyleknap: I think we should call this TestIOStreamingWriteTask. Probably forgot to change on copy and paste.

kyleknap:
Approach looks fine. There is one edge case that you may need to worry about with the deferred queue and retries and I had some smaller comments as well.

s3transfer/download.py
jamesls: Ideally I wanted to extract the `_submit_task` into a separate class so other things besides a Task could submit things.  Would giving the transfer coordinator the ability to submit tasks make sense?  We'd then update `_submit_task` to just go through the transfer coordinator.

s3transfer/download.py
jamesls: Ok I'll update.

s3transfer/download.py
jamesls: What I had in mind was:

* Ignore/discard any writes that are less than the `_next_offset`.
* Ignore any duplicate entries we receive for pending writes.  Based on the atomicity of `request_writes` and the guaranteed order of queued IO within a given GetObjectTask I don't actually think this is possible in practice, but I'd rather not have this class rely on that assumption for correctness.

I think that in order to address the second point I need to add an auxiliary structure to keep track of what offsets are pending (probably just a set of integers of offsets).

s3transfer/download.py
kyleknap: We talked about this. I would like this.

s3transfer/download.py
kyleknap: Yeah that makes sense for both.

jamesls:
@kyleknap all feedback should be incorporated now.  The bug fix for retried IO is [here](https://github.com/boto/s3transfer/pull/25/commits/83a1d45b475cd0b1779297c15032c7be8e1f7ed0).  Few things worth noting:

* Just used `hasattr(fileobj, 'write')` instead of adding a compat `writable` function.
* Didn't add the ability to submit tasks through the transfer coordinator.  Given we'd consolidate all task submission, I'd imagine we should address that in a separate PR.

kyleknap:
Sounds good. Updates look good. :ship:

PR #26, opened 2016-07-07
Add sliding window semaphore

jamesls:
The idea is to use this as a primitive to control
access to parallel range get downloads to a non
seekable stream.  We need a sliding window to ensure
we limit the amount of memory used to buffer
part data.

In the initial version, I required you to pass in a tag name as well
as a counter (e.g tag='stream-download-foo, counter=1).  This proved
to be problematic because you never actually know what the lowest 
value will be.

cc @kyleknap @JordonPhillips

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -13,6 +13,9 @@
 import os.path
 import shutil
 import tempfile
+import threading
+import random
+import time
 
 from tests import unittest
 from tests import RecordingSubscriber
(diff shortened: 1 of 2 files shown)


tests/unit/test_utils.py
```diff
@@ -374,3 +379,230 @@ def test_callback_invoked(self):
...
+    def start_threads(self):
+        for thread in self.threads:
+            thread.start()
+
+    def test_acquire_blocks_until_release_is_called(self):
+        sem = SlidingWindowSemaphore(2)
+        sem.acquire('a', blocking=False)
+        sem.acquire('a', blocking=False)
+        def acquire():
+            # This next call to acquire will bloc
```
kyleknap: bloc --> block

kyleknap:
Looks good. Just had a comment about a spelling mistake.

PR #28, opened 2016-07-08
Check file mode in readable

### LINKED ISSUE #27: readable() is inconsistent across python versions
I was looking to copy what `readable()` does in order to add a similar `writable()` function, but it has inconsistent behavior across python versions.  `readable()` will tell you if the file was opened readable (or rather if `read()` will actually succeed), whereas the code path in python2 will simply tell you if the fileobj has a read method, without any comment on whether or not read will fail.  Consider:

```
Python 2.7.11 (default, Jan 22 2016, 08:29:18)
>>> from s3transfer.compat import readable
>>> readable(open('/tmp/asdfafgafdsfd', 'w'))
True

Python 3.5.1 (default, Jan 22 2016, 08:54:32)
>>> from s3transfer.compat import readable
>>> readable(open('/tmp/adfasfasdfd', 'w'))
False
```

I would expect us to try, within reason, to have consistent behavior across python versions.

JordonPhillips:
Python 2 doesn't have the readable function on file objects, so
to inspect their readability we need to look at the file mode.

Fixes #27 

cc @kyleknap @jamesls

### DIFF
--- a/s3transfer/compat.py
+++ b/s3transfer/compat.py
@@ -88,4 +88,7 @@ def readable(fileobj):
     if hasattr(fileobj, 'readable'):
         return fileobj.readable()
 
+    if hasattr(fileobj, 'mode'):
+        return 'r' in fileobj.mode or '+' in fileobj.mode
+
     return hasattr(fileobj, 'read')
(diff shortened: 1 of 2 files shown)


jamesls:
Given the original issue is now closed, let's go ahead and close this out.

Issue #20, opened 2016-06-29
Submission thread seems to block on unrelated tags


jamesls:
I'm working in some of the future_tag code, and I'm seeing (https://github.com/boto/s3transfer/blob/develop/s3transfer/futures.py#L328)

```python

        with self._lock:
            # Wait if there is a maximum reached
            self._wait_for_futures_to_complete_if_needed(tag)
            # Submit the task to the executor
            future = self._executor.submit(fn, *args, **kwargs)
            # Add the recently added future to the currently running futures
            self._add_future_to_currently_running(future, tag)
        return future
```

So suppose I have a max tags of `{"A": 1, "B": 1}` and sequence such as:

```
T1: submit tag A --> noblock
T1: submit tag A --> block on waiting futures
T2: submit tag B --> block on lock acquire
```

Where T1/T2 are threads and A/B are tags.  This means that even though T2 is submitting tag B which has no futures, the submission on tag B will still block on the `with self._lock` until a future for tag A completes.

kyleknap:
Fixed with this: https://github.com/boto/s3transfer/pull/31. Closing.

PR #30, opened 2016-07-12
Move submission of tasks to transfer coordinator

kyleknap:
Introduced submit() method for tasks to submit other tasks to executors
in order to ensure task submission happens under one function.

Brought up because of comments in this PR: https://github.com/boto/s3transfer/pull/25 

I also fixed some flake8 issues that I noticed...

cc @jamesls @JordonPhillips

PR #31, opened 2016-07-13
Integrating Semaphores into Executor

kyleknap:
In order to support the SlidingWindowSemaphore and fix https://github.com/boto/s3transfer/issues/20, I decided to rip out the current implementation on how we bound the executors and replace it with a semaphore approach.

The PR builds off of: https://github.com/boto/s3transfer/pull/30. You can look at that one or look at both of the PR's via this one.

The larger changes in this PR include:

* Add a unique identifier for each transfer request.
* Instead of relying on current.futures.wait to wait for futures to complete we use semaphores to bound the limits. 

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -18,6 +18,7 @@ import threading
 
 from s3transfer.compat import MAXINT
 from s3transfer.utils import FunctionContainer
+from s3transfer.utils import TaskSemaphore
 
 
 logger = logging.getLogger(__name__)
@@ -227,7 +228,7 @@ class TransferCoordinator(object):
         with self._lock:
             self._status = 'running'
 
-    def submit(self, executor, task, future_tag=None):
+    def submit(self, executor, task, tag=None):
         """Submits a task to a provided executor
 
         :type executor: s3transfer.futures.BoundedExecutor
@@ -236,8 +237,8 @@ class TransferCoordinator(object):
         :type task: s3transfer.tasks.Task
         :param task: The task to submit to the executor
 
-        :type future_tag: s3transfer.futures.FutureTag
-        :param future_tag: A future tag to associate to the task
+        :type tag: s3transfer.futures.TaskTag
+        :param tag: A tag to associate to the submitted task
 
         :rtype: concurrent.futures.Future
         :returns: A future representing the submitted task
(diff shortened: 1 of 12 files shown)


s3transfer/futures.py
```diff
@@ -78,8 +78,9 @@ def cancel(self):
 
 class TransferMeta(object):
     """Holds metadata about the TransferFuture"""
-    def __init__(self, call_args=None):
+    def __init__(self, call_args=None, id=None):
```
JordonPhillips: This shadow's the [built-in id](https://docs.python.org/3/library/functions.html#id).

s3transfer/manager.py
```diff
@@ -40,8 +41,8 @@ def __init__(self,
                  multipart_chunksize=8 * MB,
                  max_request_concurrency=10,
                  max_submission_concurrency=5,
-                 max_request_queue_size=0,
-                 max_submission_queue_size=0,
+                 max_request_queue_size=1000,
+                 max_submission_queue_size=1000,
```
JordonPhillips: Why the change here? Just to put a default bound on the memory usage?

JordonPhillips:
I don't see any major issues, though I think @jamesls should have a look.

s3transfer/manager.py
kyleknap: Yeah just to protect the users. Also since we are using semaphores, a value of zero is strangely acceptable but causes deadlocks if you try to acquire it.

s3transfer/futures.py
kyleknap: It should not be an issue because it is isolated to the class but I can update it to something more explicit.

kyleknap:
Alright I updated the code based on feedback. Let me know what you think.

s3transfer/utils.py
```diff
@@ -389,7 +389,39 @@ class NoResourcesAvailable(Exception):
...
-class SlidingWindowSemaphore(object):
+class TaskSemaphore(object):
+    def __init__(self, count):
+        """A semaphore for the purpose of limiting the number of tasks
+
+        :param count: The size of semaphore
+        """
+        self._semaphore = threading.Semaphore(count)
+
+    def acquire(self, tag, blocking=True):
```
jamesls: Probably worth documenting/commenting the reason there's params that are required but not used is because we want to be API compatible with the `SlidingWindowSemaphore`.

jamesls:
:shipit:  Looks good to me.

kyleknap:
I added the docstring. @JordonPhillips let me know if you have anymore comments.

JordonPhillips:
⛵

PR #32, opened 2016-07-13
Limit in-memory chunks for downloading to non-seekable streams

kyleknap:
This integrates these two PRs: https://github.com/boto/s3transfer/pull/31 and https://github.com/boto/s3transfer/pull/26

This is the final step to be able to limit non-seekable downloads while perserving use of bandwidth. I probably would not suggest looking at this PR first. The only reason I posted it is to give people a reference as to where I am going with the first two: https://github.com/boto/s3transfer/pull/30 and https://github.com/boto/s3transfer/pull/31. So I would recommend looking at those first.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -26,6 +26,7 @@ from botocore.vendored.requests.packages.urllib3.exceptions import \
 from s3transfer.compat import SOCKET_ERROR
 from s3transfer.compat import seekable
 from s3transfer.exceptions import RetriesExceededError
+from s3transfer.futures import IN_MEMORY_DOWNLOAD_TAG
 from s3transfer.utils import random_file_extension
 from s3transfer.utils import get_callbacks
 from s3transfer.utils import invoke_progress_callbacks
@@ -72,6 +73,16 @@ class DownloadOutputManager(object):
             otherwise returns False.
         """
         raise NotImplementedError('must implement is_compatible()')
+
+    def buffers_body_in_memory(self):
+        """Whether the body it downloads is buffered in-memory
+
+        :rtype: boolean
+        :returns: True if the body returned by the manager will be buffered in
+            memory (excluding the io queue buffer). False if the manager will
+            not directly buffer the body in memory.
+        """
+        return False
 
     def get_fileobj_for_io_writes(self, transfer_future):
         """Get file-like object to use for io writes in the io executor
@@ -192,6 +203,9 @@ class DownloadNonSeekableOutputManager(DownloadOutputManager):
     def is_compatible(cls, download_target):
         return hasattr(download_target, 'write')
 
+    def buffers_body_in_memory(self):
+        return True
+
     def get_fileobj_for_io_writes(self, transfer_future):
         return transfer_future.meta.call_args.fileobj
 
@@ -308,6 +322,9 @@ class DownloadSubmissionTask(SubmissionTask):
         # Get the needed callbacks for the task
         progress_callbacks = get_callbacks(transfer_future, 'progress')
 
+        # Get any associated tags for the get object task.
+        get_object_tag = self._get_download_task_tag(download_output_manager)
+
         # Submit the task to download the object.
         download_future = self._transfer_coordinator.submit(
             request_executor,
@@ -324,7 +341,8 @@ class DownloadSubmissionTask(SubmissionTask):
                     'download_output_manager': download_output_manager,
                     'io_chunksize': config.io_chunksize,
                 }
-            )
+            ),
+            tag=get_object_tag
         )
 
         # Send the necessary tasks to complete the download.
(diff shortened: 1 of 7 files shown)


s3transfer/manager.py
```diff
@@ -45,7 +47,8 @@ def __init__(self,
                  max_submission_queue_size=1000,
                  max_io_queue_size=1000,
                  num_download_attempts=5,
-                 max_in_memory_upload_chunks=10):
+                 max_in_memory_upload_chunks=10,
+                 max_in_memory_download_chunks=10):
```
JordonPhillips: Would customers want these to be different? If you're limiting chunks in memory, does it matter what put it there?

kyleknap: It unfortunately does because ``max_in_memory_upload_chunks`` have a difference performance implication than ``max_in_memory_downloads``. 

Currently if I want to combine the two, the two would have to share the ``SlidingWindowSemaphore`` which may be slower than a regular semaphore because the release of the count of a semaphore is dependent on the lowest sequence number in-progress to finish and that may result in threads not being able to make requests until the lowest sequence number to finish.

With the regular semaphore, the sequence number has no affect on the count when the semaphore is released. So no matter what you can guarantee that if the semaphore is completely exhausted it is because there are ``max_in_memory_upload_chunk`` number of threads completing requests, which that guarantee cannot be made for streaming downloads. In the end, the sharing of the``SlidingWindowSemaphore`` could potentially slow down the in memory uploads.

I think I am on the stance that we re-address these parameters before ``s3transfer`` goes GA to see if we can combine the two (as I do think it is a little strange), but for now since it is not publicly exposed and it requires an additional feature to be able to support a ``max_in_memory_chunks`` parameter effectively.

How does that sound? Does that make sense?

kyleknap:
Alright I updated this PR with all of the other PRs getting merged into the develop branch. So it is much more manageable to review.

tests/__init__.py
```diff
@@ -435,3 +443,26 @@ def write(self, b):
...
+
+    def readable(self):
+        return False
+
+    def write(self, b):
+        self._fileobj.write(b)
+
+    def read(self, n=-1):
+        # This is needed because python will not always return the correct
+        # kind of error even though returnable returns False.
```
JordonPhillips: s/returnable/readable

Also, are you sure this is true for `read`?

s3transfer/download.py
```diff
@@ -192,6 +203,9 @@ def __init__(self, osutil, transfer_coordinator, io_executor,
     def is_compatible(cls, download_target):
         return hasattr(download_target, 'write')
 
+    def buffers_body_in_memory(self):
```
jamesls: Can you talk about the thought/motivation behind adding a new method to the manager?  This seems to be leaking implementation details (i.e. why does it matter to a caller how the manager decides to manager the internals of the body?)

Given the point of this method is solely (for now) to figure out the appropriate tags for the task, how is this approach better than just directly asking the download manager about any tags to associate with the task?  e.g `download_manager.get_tags()` or `download_manager.get_task_tags()`, etc.

jamesls:
Seems reasonable, had a design question about the new method on the managers.

s3transfer/download.py
kyleknap: Yeah I am good with calling it ``get_task`` or ``get_get_object_task`` tags. I only did it the way I did to make it similar in functionality to the upload input managers which asks if the body gets buffered in memory. But since this is not dependent on if it is a multipart upload or not a multipart upload, I am fine with changing it.

kyleknap:
@jamesls @JordonPhillips 

Updated based on feedback. Let me know what you think.

jamesls:
:shipit:

JordonPhillips:
```
___________________________________________________________________
||     *                            *                    ((   *  ||
||        *                 *                *            ~      ||
||                ___.                          *          *     ||
||       *    ___.\__|.__.           *                           ||
||            \__|. .| \_|.                                      ||
||            . X|___|___| .                         *           ||
||          .__/_||____ ||__.            *                /\     ||
||  *     .  |/|____ |_\|_ |/ _                          /  \    ||
||        \ _/ |_X__\|_  |\||~,~{                       /    \   ||
||         \/\ |/|    |_ |/:|`X'{                   _ _/      \__||
||          \ \/ |___ |_\|_.|~~~                   /    . .. . ..||
||         _|X/\ |___\|_ :| |_.                  - .......... . .||
||         | __\_:____ |  ||o-|            ___/........ . . .. ..||
||         |/_-|-_|__ \|_ |/--|       ____/  . . .. . . .. ... . ||
|| ........:| -|- o-o\_:_\|o-/:....../...........................||
|| ._._._._._\=\====o==o==o=/:.._._._._._._._._._._._._._._._._._||
|| _._._._._._\_\ ._._._._.:._._._._._._._._._._._._._.P_._._._._||
|| ._._._._._._._._._._._._._._._._._._._._._._._._._._._._._._._||
||---------------------------------------------------------------||
```

PR #33, opened 2016-07-20
Fix a memory leak from future association 

kyleknap:
We never disassociated the futures until the entire a particular transfer request actually finished. This is a **big** issue for downloads because the io tasks store the data in the task and since their futures never lose reference because of the association that memory never gets cleared up. You do not really notice this for smaller files (< 1GB) but I discovered this when I started to test large downloads ~10GB.

Note that this based off of all of the previous PR's so it may take a while to get this merged, but really only the last commit needs to be reviewed for this specific PR.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -114,7 +114,7 @@ class TransferCoordinator(object):
         self._status = 'queued'
         self._result = None
         self._exception = None
-        self._associated_futures = []
+        self._associated_futures = set()
         self._failure_cleanups = []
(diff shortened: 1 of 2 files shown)


kyleknap:
@jamesls @JordonPhillips 
I realized that this PR did not need to be based on top of the semaphore integration so I redid this PR to go against the develop branch making it so this PR does not get blocked anymore as it is pretty important for performance reasons.

jamesls:
:shipit:

PR #34, opened 2016-07-20
Add scripts for performance testing

kyleknap:
The following scripts were added:

* benchmark - Benchmark time, memory, cpu, network in/out for any script
* upload-file - Uploads a single file with s3transfer
* download-file - Downloads a single file with s3transfer
* benchmark-upload - Benchmarks the upload of a single file with s3transfer
* benchmark-download - Benchmarks the download of a single file with s3transfer

On a side note, I would much rather have my other PR's looked at than this one and have this one deprioritized. I figured that since I have these performance scripts that I have been using for a while now that I would push them up for other people to use because I have found them really handy.

cc @jamesls @JordonPhillips


scripts/performance/benchmark
```diff
@@ -0,0 +1,122 @@
...
+import psutil
+
+
+# Determine the interface to track network IO depending on the platform.
+if sys.platform.startswith('linux'):
+    INTERFACE = 'eth0'
+elif sys.platform == 'darwin':
+    INTERFACE = 'en0'
+else:
+    raise RuntimeError('Script cannot be run on %s' % sys.platform)
```
jamesls: Just curious, this script could eventually be made to run on windows (psutils exposes everything we need right?)

scripts/performance/benchmark
```diff
@@ -0,0 +1,122 @@
...
+            recv_delta = current_net.bytes_recv - previous_net.bytes_recv
+
+            # Determine the lapsed time to determine the network io rate.
+            current_time = time.time()
+            dt = current_time - previous_time
+            previous_time = current_time
+            sent_rate = sent_delta / dt
+            recv_rate = recv_delta / dt
+
+            t = time.time()
```
jamesls: Can't you just use `current_time` here?

scripts/performance/benchmark
```diff
@@ -0,0 +1,122 @@
...
+            f.flush()
+    sys.exit(0)
+
+
+def main():
+    parser = argparse.ArgumentParser(usage=__doc__)
+    parser.add_argument(
+        'script', help='The script to run for benchmarking')
+    parser.add_argument(
+        '--data-interval', default=0.1, type=float,
```
jamesls: Did setting the interval to something higher (say 1 second) change any of the CPU percentages?  I wonder if collecting so frequently starts to introduce additional load on the system that could influence the numbers.

scripts/performance/benchmark
```diff
@@ -0,0 +1,122 @@
...
+        gone, alive = psutil.wait_procs(procs, timeout=1)
+        for child in alive:
+            child.kill()
+        sys.exit(0)
+
+
+def run_script(args):
+    # Run the script in a separate process.
+    script_args = args.script.split(' ')
+    return subprocess.Popen(script_args)
```
jamesls: Why not just use `shell=True`?  That way a user can use pipes and stdout/stdin right?  Seems like it's not an issue since you're recursing through children processes.

jamesls:
Looks good, had some high level comments:

* I prefer having all the requirements in the top level directory, so you can see which requirements you may need to install.  What about moving this to a top level `requirements-perf.txt` or possibly a more general `requirements-dev.txt`?  The latter file could also include `requirements-test.txt`.

* Adding a `utils` module in the perf script requires the user needs to ensure that module is importable, so you either need to mess around with your python path or you need to run it from that directory.  Do you imagine we'd add a lot more utility functions?  It would be nice if all the scripts were self contained.

scripts/performance/benchmark
JordonPhillips: Yeah, the interface would be different though. It could actually be different if the user has a more complicated setup, such as if they're using en1 by default or if their interface has a different naming scheme.

scripts/performance/benchmark
JordonPhillips: The setting in the script I had been using was 1 second. A lower setting can be useful for quicker running commands, but yeah it does add a bit of overhead. I don't know that it would be that significant though. It'd be interesting to see the numbers.

scripts/performance/benchmark
kyleknap: Yep this something we can look to do. I just have not done really any testing with these scripts on windows.

scripts/performance/benchmark
kyleknap: Yeah I will update.

scripts/performance/benchmark
kyleknap: It was not really noticeable on the EC2 instance that I was running on, but I can do some quick testing if you want. The main issue is that what I have found CPU measurements have a relatively high standard deviation when compared to measurements of time and memory so it may take a bit more work to tell how much of an effect a change of the data interval will have on it.

scripts/performance/benchmark
kyleknap: I think it should work. Let me try it out to see if there is any issues with doing that.

kyleknap:
I am fine with the general comments as well. Seems reasonable.

scripts/performance/benchmark
jamesls: I'd be inclined to stick with @JordonPhillips 1 second interval unless we really need .1 second granularity.  It can always be overriden with the `--data-interval` argument anyways.

jamesls:
Flipping to incorporating feedback for now.

scripts/performance/benchmark
kyleknap: The CPU usage increased about 10 to 25 % for process lasting 15 to 90 seconds. So I would be fine with upping the interval.

kyleknap:
@jamesls @JordonPhillips I think I addressed everything and I did some testing on the difference between 0.1 and 1 intervals for CPU usage (there was a noticeable change in CPU performance so I updated the default). Let me know what you think.

scripts/performance/benchmark
```diff
@@ -0,0 +1,124 @@
...
+    p = psutil.Process(pid)
+    previous_net = psutil.net_io_counters(pernic=True)[INTERFACE]
+    previous_time = time.time()
+
+    with open(output_file, 'w') as f:
+        while p.is_running():
+            if p.status() == psutil.STATUS_ZOMBIE:
+                p.kill()
+                break
+            # Sleep for the data interval.
```
jamesls: FWIW, these comments are distracting.  I think it's pretty clear that `time.sleep(data_internal)` is sleeping for the data_interval.

In general, I only prefer comments that explain non-obvious things about the code or *why* the code is written this particular way.

scripts/performance/benchmark
```diff
@@ -0,0 +1,124 @@
...
+"""
+Use for benchmarking performance of other scripts. Provides data about
+time, memory use, cpu usage, network in, network out about the script ran in
+the form of a csv.
+
+
+Usage
+=====
+
+NOTE: Make sure you run ``pip install -r requirements.txt`` before running.
```
jamesls: Isn't this `requirements-dev.txt` now?

scripts/performance/benchmark
```diff
@@ -0,0 +1,124 @@
...
+            previous_time = current_time
+            sent_rate = sent_delta / dt
+            recv_rate = recv_delta / dt
+
+            # Save all of the data into a CSV file.
+            f.write(','.join(str(val) for val in [
+                current_time, memory_used, cpu_percent, sent_rate,
+                recv_rate]) + '\n')
+            f.flush()
+    sys.exit(0)
```
jamesls: Sorry I missed this.  You generally shouldn't be `sys.exit()`'ing from functions.  Given the `if __name__...` block already has a `sys.exit`, you can just return the RC.

scripts/performance/benchmark-download
```diff
@@ -0,0 +1,171 @@
...
+                temp_file, args.file_type, args.s3_bucket, s3_key)
+        )
+        # Add in the output-file for benchmark if it was specified.
+        benchmark_args = ['./benchmark', download_file_script]
+        if args.output_file:
+            benchmark_args.extend(['--output-file', args.output_file])
+        # Run the benchmark script now.
+        subprocess.check_call(benchmark_args)
+    finally:
+        # Remove the scratch directory.
```
jamesls: Same thing as before.  These comments are distracting.  I think it's reasonable to assume a python person will know what `shutil.rmtree` does.

jamesls:
:shipit: Looks good.  I had an overarching stylistic comment that some of the code comments are distracting, so just something to watch out for going forward.

scripts/performance/benchmark
kyleknap: Ahh right I did not see that either when I refactored this script from another script.

kyleknap:
I updated based on @jamesls comments.

@JordonPhillips let me know if you have any more feedback.

JordonPhillips:
⛵

PR #35, opened 2016-07-20
Add ability to configure io chunksize

kyleknap:
This is very important for environments that have very fast bandwidth and slow disk writes because if the io chunksize is too small it starts to become the bottleneck when downloading files.

This will be used to help address this issue: https://github.com/boto/boto3/issues/691 as io chunksize configuration is needed to reach the CLI-level performance on those types of environments.

The reason we just do not update the default ``STREAM_CHUNK_SIZE`` to 1MB is that will totally destroy the granularity of callbacks and is not great for slow connections.

As more evidence for this change, the functional tests even needed the change as we would currently patch and return the value of ``STREAM_CHUNKS_SIZE`` in some of the functional download tests, which is a little hacky.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -323,6 +323,7 @@ class DownloadSubmissionTask(SubmissionTask):
                     'callbacks': progress_callbacks,
                     'max_attempts': config.num_download_attempts,
                     'download_output_manager': download_output_manager,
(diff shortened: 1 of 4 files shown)


JordonPhillips:
⛵

PR #38, opened 2016-08-01
Fix deadlock isssue with concurrent.futures.wait()

kyleknap:
Ran into this issue with using concurrent.futures.wait(): https://bugs.python.org/issue20319. To avoid this bug, wait() is reimplemented by iterating over futures and calling result().

We ran into this issue on one of our automated test boxes, but I was able to reproduce this on a Windows EC2 instance that I have running and can confirm this PR no longer results in the deadlock.

I added no tests because they were already tests covering the use cases involving ``pending_main_kwargs`` which is where the ``wait()`` is used: https://github.com/boto/s3transfer/blob/develop/tests/unit/test_tasks.py#L431-L581 and for the SubmissionTask: https://github.com/boto/s3transfer/blob/develop/tests/unit/test_tasks.py#L81

I looked into adding a stress unit test: https://github.com/kyleknap/s3transfer/commit/4b743ad00a6e3663b3610ab086ffc277b4f60c76 but that was not reliably causing the repo case unless I created like 10,000 tasks which then caused it to hang like 1/5 times and took like over 10 seconds which was not ideal.

So I stuck with adding performance scripts for detecting hangs.

cc @jamesls @JordonPhillips


jamesls:
Can we move the script you added out of `scripts/performance` into either `scripts/` or `scripts/stress` or something else?  This isn't really a perf test, so I don't think it makes sense in `performance/`.

Otherwise, :shipit:

kyleknap:
Yeah I can move it.

PR #40, opened 2016-08-01
Add support for scripts that spin up processes

kyleknap:
Before was not able to accurately track aws cli streaming operations relative to memory usage and cpu usage because the streaming tasks spin off processes in piping inputs and outputs. Now with this PR it is possible to track these numbers.

cc @jamesls @JordonPhillips

### DIFF
--- a/scripts/performance/benchmark
+++ b/scripts/performance/benchmark
@@ -73,14 +73,11 @@ def run_benchmark(pid, output_file, data_interval):
                 p.kill()
                 break
             time.sleep(data_interval)
-
-            # Sometimes processes can spin up their own processes. We want
-            # to track these.
-            processes_to_track = [p] + p.children(recursive=True)
+            process_to_measure = _get_underlying_python_process(p)
             try:
                 # Collect the memory and cpu usage.
-                memory_used = _get_memory_used(processes_to_track)
-                cpu_percent = _get_cpu_used(processes_to_track)
+                memory_used = process_to_measure.memory_info().rss
+                cpu_percent = process_to_measure.cpu_percent()
                 current_net = psutil.net_io_counters(pernic=True)[INTERFACE]
             except psutil.AccessDenied:
                 # Trying to get process information from a closed process will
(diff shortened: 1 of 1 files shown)


jamesls:
Which streaming tasks spin off processes?  I wouldn't expect that to be the case. Could you add more info?

I'd prefer to just somehow track the actual process we care about rather than aggregate all child processes recursively.

kyleknap:
So I was doing something like this:
```
./s3transfer/scripts/performance/benchmark "cat 16KB | aws s3 cp - s3://mybucketfoo" --output-file temp.csv --data-interval 0.1
```
And that was getting me values like 2.3 MB memory usage and 0% CPU usage which is not correct at all.

Is there a better way to measure this than looking at all of the child processes? I am not sure if it is easy to determine what process that we actually care about when the script is relatively general.

jamesls:
Cat's unnecessary.  You can use `<` and `>` to pipe from/to stdin/stdout, respectively.

Does that work for you?

kyleknap:
No unfortunately. I switched it to:
```
./s3transfer/scripts/performance/benchmark "aws s3 cp - s3://mybucketfoo < 16KB" --output-file temp.csv --data-interval 0.1
```
and I was getting no readings for memory usage or disk usage. Only, the changes in this PR allow me to get the values.

jamesls:
I think that's just because of the `shell=True`.  This should actually affect even non streaming transfers.  It's actually running `/bin/sh` and passing the `aws ...` command as an argument.  The process tree looks like:

```
 |       \-+= 72362 jamesls python scripts/performance/benchmark aws s3 cp - s3://bucket/ < /tmp/largefile --output-file temp.csv --data-interval 1.0
 |         \-+- 72363 jamesls /bin/sh -c aws s3 cp - s3://bucket < /tmp/largefile
 |           \--- 72364 jamesls /usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python /usr/local/bin/aws s3 cp - s3://bucket
```

I'd suggest:

* If the first child process is `/bin/sh` (the default from subprocess.Popen with `shell=True`), then look at the grandchild process, and pick the first entry.

kyleknap:
I can do that then.

kyleknap:
@jamesls 

I updated it to look to see if there is any child processes and use the first python labeled process found. If there is no child processes, then it uses the original process passed in. The streaming and nonstreaming case work with this.

scripts/performance/benchmark
```diff
@@ -105,6 +103,17 @@ def run_benchmark(pid, output_file, data_interval):
     return 0
 
 
+def _get_underlying_python_process(process):
+    # For some scripts such as the streaming CLI commands, the process is
+    # nested under a shell script that does not account for the python process.
+    # We want to always be measuring the python process.
+    children = process.children(recursive=True)
+    for child_process in children:
+        if 'python' in child_process.name():
```
jamesls: This should be a case insensitive search.  If you look at my process tree in my original comment I have:

```
 72364 jamesls /usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python /usr/local/bin/aws s3 cp - s3://bucket
```

kyleknap:
Alright. I updated to normalize to lowercase. Let me know what you think.

jamesls:
:shipit: Thanks.

PR #42, opened 2016-08-02
Fix timing issue with unit test

kyleknap:
Removed the use of time.sleep and now rely on threading.Event. The timing issue is that the SuccessTask had a chance on never being executed because the ExceptionSubmissionTask happens before SuccessTask is ran.

cc @jamesls @JordonPhillips


kyleknap:
Alright tests are now passing again. There was a test that had a similar problem that got exposed when I wiped out the ``time.sleep``.

JordonPhillips:
⛵

jamesls:
:shipit:

PR #43, opened 2016-08-08
Add CountCallbackInvoker

kyleknap:
The abstraction of collecting all GetObjectTasks and waiting on all of the
futures introduced a small, but linear memory leak that could be seen in
really large downloads (greater than 10GB). Instead, we no longer hold
reference to all downloads and use a running count to track when all IOTasks
have been submitted.

cc @jamesls @JordonPhillips


jamesls:
:shipit:

PR #45, opened 2016-08-16
Update allowed parameters

kyleknap:
GrantWriteACL should be GrantWriteACP and needed to include
WebsiteRedirectLocation as valid parameter for uploads and copies.

Here is the documentation for all of the necessary operations:
https://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_object
https://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.copy_object
https://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.complete_multipart_upload

Noticed this when I was working on porting ``s3transfer`` into the CLI.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/manager.py
+++ b/s3transfer/manager.py
@@ -162,7 +162,7 @@ class TransferManager(object):
         'GrantFullControl',
         'GrantRead',
         'GrantReadACP',
-        'GrantWriteACL',
+        'GrantWriteACP',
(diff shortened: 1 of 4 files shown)


jamesls:
:shipit:

PR #46, opened 2016-08-23
Made multiple improvements to summarize script

kyleknap:
The following features were added:

   * Ability to summarize more than one file
   * Calculate standard deviation
   * Print summary in table

cc @jamesls @JordonPhillips

### DIFF
--- a/scripts/performance/summarize
+++ b/scripts/performance/summarize
@@ -12,19 +12,20 @@ Run this script with::
 
 And that should output::
 
-    +-----------------------+----------+----------------------+
-    | Metric                | Mean     | Standard Deviation   |
-    +=======================+==========+======================+
-    | Total Time (seconds)  | 1.200    | 0.0                  |
-    +-----------------------+----------+----------------------+
-    | Maximum Memory        | 42.3 MiB | 0 Bytes              |
-    +-----------------------+----------+----------------------+
-    | Maximum CPU (percent) | 88.1     | 0.0                  |
-    +-----------------------+----------+----------------------+
-    | Average Memory        | 33.9 MiB | 0 Bytes              |
-    +-----------------------+----------+----------------------+
-    | Average CPU (percent) | 30.5     | 0.0                  |
-    +-----------------------+----------+----------------------+
+    +------------------------+----------+----------------------+
+    | Metric over 1 run(s)   | Mean     | Standard Deviation   |
+    +========================+==========+======================+
+    | Total Time (seconds)   | 1.200    | 0.0                  |
+    +------------------------+----------+----------------------+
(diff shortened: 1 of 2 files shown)


JordonPhillips:
This doesn't gracefully handle an empty file:

```
Traceback (most recent call last):
  File "s3transfer/scripts/performance/summarize", line 248, in <module>
    main()
  File "s3transfer/scripts/performance/summarize", line 244, in main
    summarizer.process(args)
  File "s3transfer/scripts/performance/summarize", line 177, in process
    self.process_individual_file(benchmark_file)
  File "s3transfer/scripts/performance/summarize", line 185, in process_individual_file
    self._end_time = self._get_time(row)
UnboundLocalError: local variable 'row' referenced before assignment
```

scripts/performance/summarize
```diff
@@ -12,14 +12,45 @@ Run this script with::
 
 And that should output::
 
-    Total time: 1.810 seconds
-    Max memory: 114.0 MiB
-    Max cpu: 208.3 percent
-    Average memory: 67.3 MiB
-    Average cpu: 140.5 percent
+    +-----------------------+----------+----------------------+
+    | Metric                | Mean     | Standard Deviation   |
```
jamesls: It would be nice somewhere in the output to mention how many files are being summarized.

scripts/performance/summarize
```diff
@@ -129,8 +232,13 @@ class Summarizer(object):
 def main():
     parser = argparse.ArgumentParser(usage=__doc__)
     parser.add_argument(
-        '-f', '--benchmark-file', required=True,
-        help='The CSV output file from the benchmark script.')
+        '-f', '--benchmark-files', required=True, nargs='+',
```
jamesls: Given benchmark files are the main thing you need for this script to work, it seems more idiomatic CLI to just have these be args rather than options.

jamesls:
Aside from @JordonPhillips's comment, looks good to me.

scripts/performance/summarize
kyleknap: Yeah that would be helpful

scripts/performance/summarize
kyleknap: Yeah I would be fine with this. @JordonPhillips does this affect any the performance work you have been doing if I made that change?

JordonPhillips: No, the summary scripts aren't being used automatically.

kyleknap:
Alright updated based on feedback. Should be good to look at again.

jamesls:
:shipit:

JordonPhillips:
⛵

PR #47, opened 2016-08-24
Fix issue when uploading a previously seeked file

kyleknap:
The current position was not being taken into account when providing
the full file size thus the ReadFileChunk wrapper was using an improper
size when calculating length and how much more is needed to be read. Note
that this does not affect the multipart uploads because the original file
object is not used and instead data is chunked out to separate file objects
from the starting position, but a sanity test was added just in case.

Fixes https://github.com/boto/boto3/issues/784

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/upload.py
+++ b/s3transfer/upload.py
@@ -190,8 +190,8 @@ class UploadFilenameInputManager(UploadInputManager):
 
     def get_put_object_body(self, transfer_future):
         # Get a file-like object for the given input
-        fileobj = self._get_put_object_fileobj(
-            transfer_future.meta.call_args.fileobj)
+        fileobj, full_size = self._get_put_object_fileobj_with_full_size(
+            transfer_future)
 
         # Wrap fileobj with interrupt reader that will quickly cancel
         # uploads if needed instead of having to wait for the socket
@@ -203,7 +203,7 @@ class UploadFilenameInputManager(UploadInputManager):
         # Return the file-like object wrapped into a ReadFileChunk to get
         # progress.
         return self._osutil.open_file_chunk_reader_from_fileobj(
-            fileobj=fileobj, chunk_size=size, full_file_size=size,
+            fileobj=fileobj, chunk_size=size, full_file_size=full_size,
             callbacks=callbacks)
 
     def yield_upload_part_bodies(self, transfer_future, config):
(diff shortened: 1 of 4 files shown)


pisymbol:
Ok, thanks kyle! I will give it a whirl.

s3transfer/upload.py
```diff
@@ -287,8 +289,12 @@ def _get_upload_part_fileobj_with_full_size(self, fileobj, **kwargs):
         # source. So we must treat it as its own standalone file.
         return six.BytesIO(data), len(data)
 
-    def _get_put_object_fileobj(self, fileobj):
-        return fileobj
+    def _get_put_object_fileobj_with_full_size(self, transfer_future):
+        fileobj = transfer_future.meta.call_args.fileobj
+        # The current position needs to be taken into account when retrieving
+        # the full size of the file.
+        size = fileobj.tell() + transfer_future.meta.size
```
jamesls: I might be missing something, but this seems like the wrong level of abstraction here.  Why isn't this behavior rolled into `provide_transfer_size`?  Wouldn't this mean that `transfer_future.meta.size` is actually wrong because it doesn't take `fileobj.tell()` into account?

kyleknap: Yeah so ``transfer_future.meta.size()`` is always equal to how many bytes we expect to transfer over the wire. So if we are given a file of size 10 and the file is at position 6 of the file. We expect to transfer 4 bytes and that is what the ``transfer_future.meta.size`` value should be.

With how the ReadFileChunk works, it requires the full file size to determine how many bytes to return with basically the calculation:
```py
def read(self, amt):
   amt_to_read = min(amt, self.full_size - self.current_position)
   self.underlying_fileobj.read(amt_to_read)
```
So if we do not account for the full size of the file (``expected_size_to_transfer (aka transfer_future.meta.size) + current_pos``) we will get incorrect numbers. So if we take the previous example, if the file is at position 6 and the actual total file size is 10 and instead of using 10 we use ``transfer_future.meta.size`` which is 4, the read file chunk will calculate ``self.full_size - self.current_position`` to be -1 which is wrong and will use -1 to make the read because it is the minimum. 

Does this make sense?

jamesls: Yeah that makes sense, thanks for clarifying.

I wish there was a better way to abstract out this detail, but I don't know if it makes sense to move into `ReadFileChunk`.

jamesls:
:shipit:

s3transfer/upload.py
kyleknap: Yeah I am not sure. I find ``ReadFileChunk`` a little confusing at times as I already feel like it is doing a bit too much, but I have been hesitant to try to change it too much since it was sort of exposed through OSUtils in boto3 but was not considered a "public interface".

PR #48, opened 2016-08-30
Add delete() method to the transfer manager

jamesls:
The one thing worth noting is that `on_progress` callbacks are not invoked when an object is deleted.  A case could be made to invoke `on_progress` with the size of the object being deleted, but you'd have to make a `head_object` call to retrieve the size, and it doesn't seem worth the overhead just to get `on_progress` callbacks working for `delete`.  I added a note in the docstring to make this clear.

cc @kyleknap @JordonPhillips

### DIFF
--- a/s3transfer/delete.py
+++ b/s3transfer/delete.py
@@ -1,0 +1,73 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from s3transfer.tasks import Task
+from s3transfer.tasks import SubmissionTask
+
+
+class DeleteSubmissionTask(SubmissionTask):
+    """Task for submitting tasks to execute an object deletion."""
+
+    def _submit(self, client, config, osutil, request_executor,
+                transfer_future):
+        """
+        :param client: The client associated with the transfer manager
+
+        :type config: s3transfer.manager.TransferConfig
+        :param config: The transfer config associated with the transfer
+            manager
+
+        :type osutil: s3transfer.utils.OSUtil
+        :param osutil: The os utility associated to the transfer manager
+
+        :type request_executor: s3transfer.futures.BoundedExecutor
+        :param request_executor: The request executor associated with the
+            transfer manager
+
+        :type transfer_future: s3transfer.futures.TransferFuture
+        :param transfer_future: The transfer future associated with the
+            transfer request that tasks are being submitted for
+        """
+        call_args = transfer_future.meta.call_args
+
+        self._transfer_coordinator.submit(
+            request_executor,
+            DeleteObjectTask(
+                transfer_coordinator=self._transfer_coordinator,
+                main_kwargs={
+                    'client': client,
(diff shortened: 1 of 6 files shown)


s3transfer/delete.py
```diff
@@ -0,0 +1,73 @@
...
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from s3transfer.tasks import Task
+from s3transfer.tasks import SubmissionTask
+
+
+class DeleteSubmissionTask(SubmissionTask):
+    """Task for submitting tasks to execute an object deletion."""
+
+    def _submit(self, client, config, osutil, request_executor,
```
kyleknap: If config and osutils are not used, you can just make them apart of ``**kwargs``. ``_submit()`` is defined such that any arbitrary kwargs can be passed in: https://github.com/boto/s3transfer/blob/develop/s3transfer/tasks.py#L270

tests/functional/test_delete.py
```diff
@@ -0,0 +1,59 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from tests import BaseGeneralInterfaceTest
+from s3transfer.manager import TransferManager
+
+
+class TestDeleteObject(BaseGeneralInterfaceTest):
```
kyleknap: I do not think anything needs to change here, but you had asked me earlier about this so I figured to make a note here. For the functional test, I usually add a general test where I take the default stubbed responses add an extra argument and make sure that gets passed in correctly, but it looks like you already have that in the unit test so it does not really need a functional test. The functional test is only nice because it guarantees that that specific test will always be there as interfaces can come and go with unit tests.

tests/functional/test_delete.py
```diff
@@ -0,0 +1,59 @@
...
+            [{'method': 'put_object', 'service_response': {}}]
+        """
+        return [{
+            'method': 'delete_object',
+            'service_response': {},
+            'expected_params': {'Bucket': self.bucket, 'Key': self.key},
+        }]
+
+    def create_expected_progress_callback_info(self):
+        return []
```
kyleknap: So I should have put this in the BaseGeneralInterfaceTest but I think we should check to make sure that the ALLOWED_DELETE_ARGS are correct. Here is what we do for copies: https://github.com/boto/s3transfer/blob/develop/tests/functional/test_copy.py#L249

tests/integration/test_delete.py
```diff
@@ -0,0 +1,40 @@
...
+
+        self.assertFalse(self.object_exists(key_name))
+
+    def test_error_when_object_does_not_exist(self):
+        transfer_manager = self.create_transfer_manager()
+        future = transfer_manager.delete(bucket=self.bucket_name,
+                                         key='does_not_exist')
+        # The underlying DeleteObject API doesn't return any indication
+        # if you try to delete an object that does not exist.
+        result = future.result()
```
kyleknap: Wait so if you try to delete something that does not exist you do not get an error? If the ``delete_object`` call does throw an error, ``result()`` should be raising it. If ``delete_object`` does not throw an error, I am not entire sure of the usefulness of this test.

kyleknap:
It looks good. Pretty straightforward. I just had some small comments.

tests/integration/test_delete.py
jamesls: You don't get an error from S3 is you delete an object that does not exist (that you otherwise have had access to).  I initially started writing the test intending to verify it propagated the exception, but realized that S3 doesn't return an error.

I've removed the test.

jamesls:
@kyleknap Should be good to look at again.

kyleknap:
Thanks. Looks good. :ship:

JordonPhillips:
Looks good!⛵

PR #49, opened 2016-08-31
Adjust chunksize if necessary

JordonPhillips:
This will adjust the chunksize of an upload or copy if it is too small, too
large, or if it looks like max parts will be exceeded. This is unnecessary
for downloads since there are no such restrictions for ranged downloading.

cc @kyleknap @jamesls

### DIFF
--- a/s3transfer/utils.py
+++ b/s3transfer/utils.py
@@ -13,6 +13,7 @@
 import random
 import time
 import functools
+import math
 import os
 import string
 import logging
@@ -22,6 +23,12 @@ from collections import defaultdict
 from s3transfer.compat import rename_file
 
 
+MAX_PARTS = 10000
+# The maximum file size you can upload via S3 per request.
+# See: http://docs.aws.amazon.com/AmazonS3/latest/dev/UploadingObjects.html
+# and: http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
+MAX_SINGLE_UPLOAD_SIZE = 5 * (1024 ** 3)
+MIN_UPLOAD_CHUNKSIZE = 5 * (1024 ** 2)
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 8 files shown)


jamesls:
What do you all think about not making the adjust chunk size "global"?  That's one of the quirks I remember from the CLI/boto3 that this meant you had to update every single test and use "realistic" sized data in unit tests when they don't really matter (for those particular tests).

What if these objects instead took a "part size adjuster".  We could then have the tests use the "noop" version for unit tests that don't actually care and in the "real" code we could pass the actual chunk size adjuster.

Also, just a heads up, the travis build failed in case you haven't seen it yet.

jamesls:
Just wanted to capture offline conversation I had with @kyleknap .  Let's go ahead and update this to be configurable.  I think having a this validation be global was reasonable for the CLI, but I think for this library it makes sense to allow this to be configurable.

@kyleknap let me know if I missed anything.

JordonPhillips:
I could see adding this to our transfer config or as an optional parameter to `copy` and `upload`, but it seems odd to change our public interface to make testing marginally simpler.

JordonPhillips:
As we discussed offline, there's some consideration on how we would want to open this up in the future. In the meantime, we'll settle for leaving it open to configuration by allowing an adjuster to be provided with `main_kwargs`.

s3transfer/utils.py
```diff
@@ -567,3 +574,57 @@ def release(self, tag, acquire_token):
...
+
+
+class ChunksizeAdjuster(object):
+    def __init__(self, max_size=MAX_SINGLE_UPLOAD_SIZE,
+                 min_size=MIN_UPLOAD_CHUNKSIZE, max_parts=MAX_PARTS):
+        self.max_size = max_size
+        self.min_size = min_size
+        self.max_parts = max_parts
+
+    def adjust_chunksize(self, current_chunksize, file_size=None):
```
jamesls: Why do we allow file_size to be `None`?  Do we actually need that behavior?

jamesls:
I'll probably defer to @kyleknap on this, but after seeing this version, I'm a little hesitant about this new concept of default values in task submission.  These are essentially params that will never be invoked by "real" code (i.e through the TransferManager public API).

If we're not going to offer a public API right now, I'm not sure if there's much benefit in plumbing this in right now.

I suppose that puts me at -0 but I can be convinced.

s3transfer/utils.py
JordonPhillips: Yes. We don't know the file size when uploading non-seekable files.

kyleknap: That makes sense. It would be good to document what ``None`` implies here though.

tests/unit/test_utils.py
```diff
@@ -683,3 +685,34 @@ def acquire():
...
+        new_size = self.adjuster.adjust_chunksize(chunksize, file_size)
+        self.assertEqual(new_size, MAX_SINGLE_UPLOAD_SIZE)
+
+    def test_chunksize_too_small(self):
+        chunksize = 7 * (1024 ** 2)
+        file_size = 5 * (1024 ** 4)
+        # If we try to upload a 5TB file, we'll need to use 896MB part
+        # sizes.
+        new_size = self.adjuster.adjust_chunksize(chunksize, file_size)
+        self.assertEqual(new_size, 896 * (1024 ** 2))
```
kyleknap: We should also test for when the file size is ``None``.

tests/unit/test_utils.py
```diff
@@ -683,3 +685,34 @@ def acquire():
...
+        file_size = MAX_SINGLE_UPLOAD_SIZE * 2
+        new_size = self.adjuster.adjust_chunksize(chunksize, file_size)
+        self.assertEqual(new_size, MAX_SINGLE_UPLOAD_SIZE)
+
+    def test_chunksize_too_small(self):
+        chunksize = 7 * (1024 ** 2)
+        file_size = 5 * (1024 ** 4)
+        # If we try to upload a 5TB file, we'll need to use 896MB part
+        # sizes.
+        new_size = self.adjuster.adjust_chunksize(chunksize, file_size)
```
kyleknap: It would also be good to make sure that the number of parts you get out of it is less than the max number of parts.

kyleknap:
Yeah I would prefer holding off on making the ``ChunksizeAdjuster`` an optional parameter you can provide to ``_submit()`` mainly because it is not really used in the code at all.

Otherwise it looks good. Just had comments on the tests.

JordonPhillips:
Updated

tests/unit/test_utils.py
```diff
@@ -683,3 +686,41 @@ def acquire():
...
+        self.assertEqual(new_size, 896 * (1024 ** 2))
+        num_parts = file_size / new_size
+        self.assertLessEqual(num_parts, MAX_PARTS)
+
+    def test_unknown_file_size(self):
```
jamesls: It would be nice to add a test when we don't know the file size and we actually make an adjustment (either above/below the thresholds).

jamesls:
Looks good to me, just had a small request for a test.

JordonPhillips:
Tests added

kyleknap:
Thanks. Looks good. :ship: Also make sure you add a changelog entry.

jamesls:
:shipit:

PR #50, opened 2016-09-08
Add aggregated progress callbacks to uploads

kyleknap:
The ``on_progress`` callbacks for uploads are too grainular (like 8KB). This is a problem for faster network environments because if you have any synchronization mechanisms in place for on_progress, it will slow down the transfer substantially because the ``read() ``is often much faster than the mechanisms that may be getting invoked in the callback. The abstraction
aggregates these smaller progress updates into a larger one to help ensure that these slower mechanisms are not called too often. The default of 256KB was chosen as it is still pretty grainular for non-multipart uploads and there is not too much of improvement past that value based off of testing.

I tested this on a couple of machines and I have seen quite a bit of improvements in the CLI integration:
1) Speed. Like I described in the paragraph above
2) Memory. When there are callback for every 8KB, it floods the result queue and the processor cannot keep up.
3) CPU. With 8KB being sent every time, it eats a lot of CPU when constantly having to fight for the lock over the result queue.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/upload.py
+++ b/s3transfer/upload.py
@@ -24,6 +24,38 @@ from s3transfer.utils import get_callbacks
 from s3transfer.utils import DeferredOpenFile, ChunksizeAdjuster
 
 
+class AggregatedProgressCallback(object):
+    def __init__(self, callback, threshold=1024 * 256):
+        """Aggregates progress updates into a single update
+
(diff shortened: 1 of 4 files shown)


jamesls:
:shipit: pretty straightforward.

PR #51, opened 2016-09-12
Add support for downloading to special files

kyleknap:
For UNIX special files like /dev/null, we cannot write to a temporary file and move it over. We need to open the file and write to it directly. Testing this with the CLI and downloading to /dev/null works fine.

Two things to note that comments would be great about:
* I decide if it is a special file name through the ``is_compatible`` methods. I decided to create a new class instead of have it built into ``DownloadFilenameOutputManager`` because it would be a good amount of branching logic based on if we are downloading to a temporary file or not. The one compensation that had to be made is that I had to add osutils as a part of the signature to ``is_compatible``, which I think is fine because it makes sense if you want to introspect the fileobj a bit more in the future to determine compatibility.
* I was having trouble unit testing all of the ``is_special_file`` cases without patching or creating these special files. There is not a great way to test them other than trying them out by hand, which I did when I first implemented it in the CLI (and this is a complete copy and paste). Any suggestions on this would be great. 

cc @jamesls @JordonPhillips

### DIFF
--- a/tests/functional/test_utils.py
+++ b/tests/functional/test_utils.py
@@ -1,0 +1,44 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import os
+import shutil
+import stat
+import socket
+import tempfile
+
+from tests import unittest
+from tests import skip_if_windows
+from s3transfer.utils import OSUtils
+
+
+@skip_if_windows('Windows does not support UNIX special files')
+class TestOSUtilsSpecialFiles(unittest.TestCase):
+    def setUp(self):
+        self.tempdir = tempfile.mkdtemp()
+        self.filename = os.path.join(self.tempdir, 'myfile')
+
+    def tearDown(self):
+        shutil.rmtree(self.tempdir)
+
+    def test_character_device(self):
+        self.assertTrue(OSUtils().is_special_file('/dev/null'))
+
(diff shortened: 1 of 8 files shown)


tests/integration/test_download.py
```diff
@@ -165,3 +166,18 @@ def test_above_threshold_for_nonseekable_fileobj(self):
...
+    def test_download_to_special_file(self):
+        transfer_manager = self.create_transfer_manager(self.config)
+        filename = self.files.create_file_with_size(
+            'foo.txt', filesize=1024 * 1024)
+        self.upload_file(filename, '1mb.txt')
+        future = transfer_manager.download(
+            self.bucket_name, '1mb.txt', '/dev/null')
+        try:
+            future.result()
+        except Exception as e:
```
kyleknap: I wonder if it is an appropriate assumption that we have permissions to write to /dev/null or if I should be creating my own special file like a fifo or socket?

jamesls: I think it's typically safe to assume `/dev/null` is writeable.

jamesls:
I think it's probably worth adding functional tests for the specific files we can create easily (fifo's, domain sockets, chr devices (/dev/null)).  Even if it's marked as skip on windows, I think that's better than nothing.

Otherwise looks good.  Updating the method signature to require `osutils` seems like a reasonable tradeoff.

kyleknap:
Sounds good to me.

kyleknap:
@jamesls I updated to add functional tests for detecting special files with OSUtils.

tests/functional/test_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def tearDown(self):
+        shutil.rmtree(self.tempdir)
+
+    def test_character_device(self):
+        self.assertTrue(OSUtils().is_special_file('/dev/null'))
+
+    def test_fifo(self):
+        mode = 0o600 | stat.S_IFIFO
+        os.mknod(self.filename, mode)
```
jamesls: Isn't there just a `os.mkfifo` you can use?

kyleknap: Yeah I can do that I think. I just copied this from the CLI tests and that is how it got set up there.

kyleknap:
@jamesls I updated to use ``os.mkfifo()``

jamesls:
:shipit:  Looks good to me.

PR #54, opened 2016-09-13
Put back original test_upload test with small content

jamesls:
This fails CI builds and uses 10MB+ of memory.  I added a TODO
about this, but for now we have to resort to patching out this value
to use a min_size of 1.

I used [this commit](https://github.com/boto/s3transfer/blob/7ae5c76c47a225218e09080abc7bb66e3732a1fa/tests/functional/test_upload.py) to put back the previous version of this file before adding in the patch.

cc @kyleknap @JordonPhillips

### DIFF
--- a/tests/functional/test_upload.py
+++ b/tests/functional/test_upload.py
@@ -14,6 +14,7 @@ import os
 import tempfile
 import shutil
 
+import mock
 from botocore.client import Config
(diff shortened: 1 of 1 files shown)


kyleknap:
Yeah not a fan of it, but the memory issues is a bigger problem. :ship:

PR #55, opened 2016-09-14
Fix hanging issue when hanging on many files

kyleknap:
This PR fixes an issue where if there is a bunch of transfer queued up in the submission executor and a Cntrl-C cancels the entire transfer, the entire transfer will hang. Essentially it boils down to the fact in the ``TransferManager.__exit__`` we will call ``TransferFutures.result()`` for all transfers that have not completed. Then if a transfer has been queued but not processed by the submission executor, the submission task's ``_main()`` will not be called because it will see that the transfer is cancelled. The problem though is that done needs to be announced to unblock ``TransferFutures.result()`` and the submission task relies on the tasks it submits to announce done. So if the submission task never submits tasks, done will never be announced.

The fix is to announce done in the submission task if we know its ``_main()`` is going to be skipped over. 

Among this change I did the following to compliment it:
* I added integration tests to catch this. I added them before implementing the fix and they were failing pretty much every time so that was good.
* I added transfer_id to the repr of ``Task`` and ``TransferCoordinator``. This will make it easier for us to determine what tasks are coordinators are related to a specific transfer and debugging for the future.
* Added some debugging in the ``wait()`` where if the transfer was hanging it will tell us what transfer it was waiting on before the user hit Cntrl-C and give us a better indication on what may be hanging in the future.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -119,7 +119,7 @@ class TransferCoordinator(object):
     """A helper class for managing TransferFuture"""
     def __init__(self, transfer_id=None):
         self.transfer_id = transfer_id
-        self._status = 'queued'
+        self._status = 'not-started'
         self._result = None
         self._exception = None
         self._associated_futures = set()
@@ -162,7 +162,9 @@ class TransferCoordinator(object):
         """The status of the TransferFuture
 
         The currently supported states are:
-            * queued - Has yet to start
+            * not-started - Has yet to start. If in this state, a transfer
+              can be canceled immediately and nothing will happen.
+            * queued - SubmissionTask is about to submit tasks
             * running - Is inprogress. In-progress as of now means that
               the SubmissionTask that runs the transfer is being executed. So
               there is no guarantee any transfer requests had been made to
@@ -171,8 +173,7 @@ class TransferCoordinator(object):
             * failed - An exception other than CancelledError was thrown
             * success - No exceptions were thrown and is done.
         """
-        with self._lock:
-            return self._status
+        return self._status
 
     def set_result(self, result):
         """Set a result for the TransferFuture
@@ -223,9 +224,18 @@ class TransferCoordinator(object):
         """Cancels the TransferFuture"""
         if not self.done():
             with self._lock:
+                should_announce_done = False
                 logger.debug('TransferCoordinator cancel() called')
                 self._exception = futures.CancelledError
+                if self._status == 'not-started':
+                    should_announce_done = True
                 self._status = 'cancelled'
+                if should_announce_done:
+                    self.announce_done()
+
+    def set_status_to_queued(self):
+        """Sets the TransferFutrue's status to running"""
+        self._status = 'queued'
 
     def set_status_to_running(self):
         """Sets the TransferFuture's status to running"""
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -195,8 +195,8 @@ class TransferCoordinator(object):
 
         Implies the TransferFuture failed.
         """
-        if not self.done():
-            with self._lock:
+        with self._lock:
+            if not self.done():
                 self._exception = exception
                 self._status = 'failed'
 
@@ -222,8 +222,8 @@ class TransferCoordinator(object):
 
     def cancel(self):
         """Cancels the TransferFuture"""
-        if not self.done():
-            with self._lock:
+        with self._lock:
+            if not self.done():
                 should_announce_done = False
                 logger.debug('TransferCoordinator cancel() called')
                 self._exception = futures.CancelledError
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -131,6 +131,10 @@ class TransferCoordinator(object):
         self._done_callbacks_lock = threading.Lock()
         self._failure_cleanups_lock = threading.Lock()
 
+    def __repr__(self):
+        return '%s(transfer_id=%s)' % (
+            self.__class__.__name__, self.transfer_id)
+
     @property
     def exception(self):
         return self._exception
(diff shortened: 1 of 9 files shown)


JordonPhillips:
The code looks good, I'll test it out and report back.

JordonPhillips:
The fix seems to work for me and the tests do pass, but when I run the tests without the fix they just hang rather than failing.

kyleknap:
Yeah I am not sure if there is a good way around preventing hangs (if actually does hang) other then spin it up in a different process. We have the same problem with the existing large file does not hang tests. The integration test that I did add are pretty much the exact same as those. I am not sure if there is a great way to handle this other than spin up an entire process and kill it if it takes to long. I am up for suggestions on this?

s3transfer/futures.py
```diff
@@ -131,6 +131,10 @@ def __init__(self, transfer_id=None):
         self._done_callbacks_lock = threading.Lock()
         self._failure_cleanups_lock = threading.Lock()
 
+    def __repr__(self):
```
JordonPhillips: This is just testing out the new review features. Nothing to see here.

s3transfer/tasks.py
```diff
@@ -279,6 +284,20 @@ def _submit(self, transfer_future, **kwargs):
...
+        # main() method.
+        if not self._transfer_coordinator.done():
+            return self._execute_main(kwargs)
+        # If the transfer coordinator signals it is done prior to the
+        # submission task running _main, the submission task will not
+        # submit any further tasks and therefore must take responsibility
+        # to announce done because there will be no tasks submitted to
+        # announce done for the transfer.
+        else:
+            self._transfer_coordinator.announce_done()
```
jamesls: This seems odd to me.  We're saying if the transfer coordinator is not done, then run `_execute_main`.  However if the transfer coordinator *is* done, then `announce_done`.  Seems like the coordinator shouldn't be allowed to be in a terminal state without something calling `announce_done` (or it's equivalent).

kyleknap: Eh I think ``done()`` is a poorly chosen name. Mainly because ``done()`` does not actually mean the transfer is done, as there is nothing left to do. ``done()`` just means that the status has been set to a terminal status, but there could be extra activities happening. The **only** thing that indicates there is nothing left to do in the transfer is the setting of that done event. The main problem is that status and this done event are set independently. In this case, I am using the fact that these are independently set to pretty much determine if it is in a "is shutting down" state.

I think given the time, I would do the following to make this logic more clear:

1) Introduce a "shutting-down" state that if an error or cancel occurred, it would be set and then shifted over to failed and cancelled once done was announced.
2) Move "success" state over from "running" when done is announced, instead of setting it when ``set_result()`` is called.
3) Remove this logic to execute main if it is not shutting down.
4) Abstract all this logic more because it is getting complicated.

Does that make sense? I could make this change but seems outside the scope of this PR, but I'd be happy to make that change. What are your thoughts/suggestions?

kyleknap:
I just improved the error message for the cntrl-c integration tests to give us a better idea of how much off from the allotted time it is off if the test fails.

kyleknap:
@jamesls Alright so I updated the PR to rely on the state transitioning of the ``TransferCoordinator`` to know when to announce done, instead of having the logic in the ``SubmissionTask``. The logic now is if ``TransferCoordinator`` is in a ``not-started`` state and the transfer gets cancelled then announce done to ensure that done actually get announced. I also started adding some state transitioning logic to help validate some of these transitions and make sure states do not get clobbered.

As to fixing the semantics of ``done()`` (i.e. done meaning at a final state and the event has been announced), this I think can be addressed in a separate PR because it is a decently sized change.

s3transfer/futures.py
```diff
@@ -219,14 +224,30 @@ def cancel(self):
         """Cancels the TransferFuture"""
         if not self.done():
             with self._lock:
```
jamesls: Should the entire method be protected by a lock?  `cancel()` would only be called in error cases anyways, and as is, I think it's still possible to have a race condition where between the if statement and lock acquisition some other thread can come along and call a method that changes the state.

s3transfer/manager.py
```diff
@@ -586,6 +586,10 @@ def wait(self):
             # If Keyboard interrupt is raised while waiting for
             # the result, then exit out of the wait and raise the
             # exception
+            if transfer_coordinator:
```
jamesls: Are you sure this is safe to assume `transfer_coordinator` is defined?  What is `KeyboardInterrupt` is raised before entering the for loop, but after the try?  Wouldn't you get an `UnboundLocalError`?

jamesls:
I know we've been following the `concurrent.futures` semantics, but given these changes, should it also be a runtime error to try to transition from a terminal state to another terminal state?  If something is finished via `set_result()`, should it be a runtime error now to come along and `cancel()` the request?

s3transfer/manager.py
kyleknap: I will add ``transfer_coordinator = None`` just before the for loop. It could happen but is improbable. You would need to hit Cntrl-C right when that for loop was entered and the list has to be empty.

s3transfer/futures.py
kyleknap: I can add the lock to now before ``self.done()``. I removed the lock around status and so before this would cause a deadlock.

kyleknap:
I think ideally it should be. I do not think I want to make this change in this pull request until I revamp the state logic and have the done states be set with the done event's.

If I do it right now, I would need to be super careful that I do not miss any edge cases (and I have a feeling there will be). For example, one edge case that I would need to account for is if I cntrl-C in the context manager of the TransferManager and a transfer has completed but not removed itself from the tracked futures, calling cancel() could propograte an error to the TransferManager, which should not be the case.

Does that make sense?

jamesls:
Seems reasonable to change the `cancel()` semantics in a separate PR.  I think I'd just like to see the other two pieces of feedback incorporated and we can call this good (locking the entire method  and setting `transfer_coordinator` to `None`.

kyleknap:
@jamesls I think I addressed the feedback. Should be good now.

PR #56, opened 2016-09-14
Add message to cancel()

kyleknap:
This builds off of this PR: https://github.com/boto/s3transfer/pull/55 so that one should get reviewed first.

The main changes that I exposed a message so that you can gain some introspection on why a transfer got cancelled. This is very useful for subscribers and calling future.result() because a general ``CancelledError`` is not too insightful and I plan to use this in the integration to the CLI to track fatal errors.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/manager.py
+++ b/s3transfer/manager.py
@@ -562,14 +562,17 @@ class TransferCoordinatorController(object):
         with self._lock:
             self._tracked_transfer_coordinators.remove(transfer_coordinator)
 
-    def cancel(self):
+    def cancel(self, msg=''):
         """Cancels all inprogress transfers
 
         This cancels the inprogress transfers by calling cancel() on all
         tracked transfer coordinators.
+
+        :params msg: The message to pass on to each transfer coordinator that
+            gets cancelled.
         """
         for transfer_coordinator in self.tracked_transfer_coordinators:
-            transfer_coordinator.cancel()
+            transfer_coordinator.cancel(msg)
 
     def wait(self):
         """Wait until there are no more inprogress transfers
--- a/s3transfer/manager.py
+++ b/s3transfer/manager.py
@@ -515,13 +515,15 @@ class TransferManager(object):
             # wrapped in a try statement because this can be interrupted
             # with a KeyboardInterrupt that needs to be caught.
             self._coordinator_controller.wait()
-        finally:
+        except KeyboardInterrupt as e:
             # If not errors were raised in the try block, the cancel should
             # have no coordinators it needs to run cancel on. If there was
             # an error raised in the try statement we want to cancel all of
             # the inflight transfers before shutting down to speed that
             # process up.
-            self._coordinator_controller.cancel()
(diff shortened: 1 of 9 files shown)


kyleknap:
@jamesls @JordonPhillips This PR is updated now with the previous dependent PR being merged. It should be good to look at.

s3transfer/manager.py
```diff
@@ -497,23 +501,29 @@ def shutdown(self, cancel=False):
...
             self._coordinator_controller.wait()
-        finally:
+        except KeyboardInterrupt as e:
             # If not errors were raised in the try block, the cancel should
             # have no coordinators it needs to run cancel on. If there was
             # an error raised in the try statement we want to cancel all of
             # the inflight transfers before shutting down to speed that
             # process up.
-            self._coordinator_controller.cancel()
+            self._coordinator_controller.cancel('keyboard interrupt')
```
kyleknap: I am not sure of a great way of testing that the message has ``'keyboard interrupt'``. What I need to do is be able to inject a KeyboardInterrupt straight into this method but there is not really anything for me to hook into because the coordinator controller is not exposed in the transfer manager and I really do not want to expose this. Does anyone have ideas on this?

s3transfer/manager.py
```diff
@@ -497,23 +501,29 @@ def shutdown(self, cancel=False):
...
+            raise e
```
JordonPhillips: You're not actually using the error object, so you can just except the class and call `raise` without any arguments to reraise it. That also has the benefit of not messing with the trace.

s3transfer/manager.py
```diff
@@ -479,15 +479,19 @@ def _register_handlers(self):
...
         cancel = False
+        cancel_msg = ''
         # If a exception was raised in the context handler, signal to cancel
         # all of the inprogress futures in the shutdown.
         if exc_type:
             cancel = True
-        self.shutdown(cancel)
+            cancel_msg = exc_value
+            if exc_type == KeyboardInterrupt:
+                cancel_msg = 'keyboard interrupt'
```
JordonPhillips: What do you think about using `repr` here when there's no message to be more generic?

kyleknap: Yeah that would be better.

s3transfer/manager.py
kyleknap: If I use the repr, than I will need reference to the error.

JordonPhillips: Yes, though using raise without arguments will still produce a neater stack trace. That's more of an opinion thing though, so I'm +0 on using `raise` over `raise e`

kyleknap: I decided in the end not too. I am just using ``raise`` with no arguments

s3transfer/manager.py
JordonPhillips: It's not great, but you could patch out the `wait` method. Not sure of a better way of doing it without exposing the controller.

kyleknap:
@JordonPhillips updated should be good to look at again.

PR #60, opened 2016-09-21
Create own version of ThreadPoolExecutor

kyleknap:
The reason this was done was that concurrent.futures version was too
complicated for what was needed and because of that there was a fair
amount of undesired deadlocks from features that we do not even care
about.

This version fo ThreadPoolExecutor is much simpler in the sense that
it spins up all of its threads at once, only has submit(), does not
concern itself with waiters, and does not register any atexit callbacks.

@JordonPhillips you can use this branch for the ``ExecutorFuture`` work
that you are doing.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -14,9 +14,11 @@ from concurrent import futures
 from collections import namedtuple
 import copy
 import logging
+import sys
 import threading
 
 from s3transfer.compat import MAXINT
+from s3transfer.compat import queue
 from s3transfer.exceptions import CancelledError
 from s3transfer.utils import FunctionContainer
 from s3transfer.utils import TaskSemaphore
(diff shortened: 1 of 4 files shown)


kyleknap:
Looks like the Python 2 port of ``concurrent.futures`` has a different interface than the Python 3 version. Mainly ``Future`` in Python 3 does not have a ``set_exception_info`` method which is needed for Python 2. I think @JordonPhillips will just build on this PR with his ``ExecutorFuture`` changes as I think we need the ``set_exception_info`` method that I use in my implementation of ``ThreadPoolExecutor``

JordonPhillips:
I like this being able to push to other pr branches

s3transfer/futures.py
```diff
@@ -422,38 +536,91 @@ def shutdown(self, wait=True):
...
 
-    def result(self):
-        return self._future.result()
+    def _invoke_callbacks(self):
+        """Invokes all the queued done callbacks."""
+        for callback in self._done_callbacks:
+            try:
+                callback()
+            except Exception:
+                logger.debug('exception calling callback for %r', self)
```
kyleknap: I realize that this is from the futures source code, but I wonder if it is worth capturing the traceback as well in the log.

s3transfer/futures.py
```diff
@@ -422,38 +536,91 @@ def shutdown(self, wait=True):
...
+
+    def result(self, timeout=None):
+        """Return the result of the future.
+
+        :type timeout: int
+        :param timeout: A number of seconds to wait for the result. A
+            TimeoutError will be raised if the future is not done in time.
+        """
+        with self._condition:
+            if self._done:
```
kyleknap: Do yo know why this is done twice: once before the wait and one after the wait:
```python
if self._done:
     return self._get_result()
```
I feel like it will work with that statement after the wait.

tests/unit/test_futures.py
```diff
@@ -480,24 +486,134 @@ def test_shutdown_no_wait(self):
...
+            raise RuntimeError()
+
+        worker = ExecutorWorker(self.task_queue)
+        worker.start()
+        future = ExecutorFuture()
+        self.task_queue.put(ExecutorWorkerTask(future, raise_runtime_error))
+        self.task_queue.put(ExecutorWorkerShutdownTask())
+        worker.join()
+        with self.assertRaises(RuntimeError):
+            future.result()
```
kyleknap: Hmm maybe I should add a test to make sure I captured the exception info as well.

tests/unit/test_futures.py
```diff
@@ -480,24 +486,134 @@ def test_shutdown_no_wait(self):
...
+            self.future.result()
+            # An exception should have been raised
+            self.fail()
+        except ValueError:
+            actual_tb = sys.exc_info()[2]
+            last_frame = traceback.extract_tb(actual_tb)[-1]
+            last_expected_frame = traceback.extract_tb(tb)[-1]
+            self.assertEqual(last_frame, last_expected_frame)
+
+    def test_done_callback(self):
```
kyleknap: I think that we should add a test to make sure an exception can be raised in the done callback and it does not affect the thread calling it (i.e. catches the exception).

s3transfer/futures.py
JordonPhillips: In case it gets past the wait due to the timeout rather than `notify` being called.

kyleknap: Ah it is because it is a condition. I was still thinking it was an event.

tests/unit/test_futures.py
```diff
@@ -34,10 +34,25 @@
...
     return args, kwargs
 
 
+def raise_exception(exception):
+    raise exception
+
+
+def get_exc_info(exception, ):
+    try:
+        raise_exception(exception)
```
JordonPhillips: I don't think the indirection is necessary here

tests/unit/test_futures.py
```diff
@@ -34,10 +34,25 @@
...
+
 def return_call_args(*args, **kwargs):
     return args, kwargs
 
 
+def raise_exception(exception):
+    raise exception
+
+
+def get_exc_info(exception, ):
```
JordonPhillips: That comma can be removed

kyleknap:
:ship: for what it is worth since I cannot approve my own PR...

jamesls:
Closing for now, not needed for now.

PR #61, opened 2016-09-22
Remove unnecessary lock

kyleknap:
The lock can be removed because it is guaranteed that the done state along with the result/error before the done event is announced. Having this lock around started causing a deadlock in the case where the main thread would call cancel, the task had not started so done gets announced, when done gets announced it runs the done callbacks, and if one of these done callbacks used the ``result()`` it would cause a deadlock where ``cancel()`` already had the lock and result() was unnecessarily requiring the lock. I also considered switching it to an ``RLock``, but it seemed like we should not be introducing if it is not needed as of now.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -216,10 +216,9 @@ class TransferCoordinator(object):
 
         # Once done waiting, raise an exception if present or return the
         # final result.
-        with self._lock:
-            if self._exception:
-                raise self._exception
-            return self._result
+        if self._exception:
+            raise self._exception
+        return self._result
 
     def cancel(self, msg='', exc_type=CancelledError):
         """Cancels the TransferFuture
--- a/tests/unit/test_futures.py
+++ b/tests/unit/test_futures.py
@@ -195,6 +195,23 @@ class TestTransferCoordinator(unittest.TestCase):
         with self.assertRaises(CancelledError):
             self.transfer_coordinator.result()
 
+    def test_cancel_can_run_done_callbacks_that_uses_result(self):
+        exceptions = []
+
+        def capture_exception(transfer_coordinator, captured_exceptions):
+            try:
+                transfer_coordinator.result()
+            except Exception as e:
+                captured_exceptions.append(e)
+
+        self.assertEqual(self.transfer_coordinator.status, 'not-started')
+        self.transfer_coordinator.add_done_callback(
+            capture_exception, self.transfer_coordinator, exceptions)
+        self.transfer_coordinator.cancel()
+
+        self.assertEqual(len(exceptions), 1)
+        self.assertIsInstance(exceptions[0], CancelledError)
+
     def test_cancel_with_message(self):
         message = 'my message'
         self.transfer_coordinator.cancel(message)


jamesls:
> The lock can be removed because it is guaranteed that the done state along with the result/error before the done event is announced.

How is this guaranteed?  Because we're calling the public APIs in a specific order?

kyleknap:
Yes it is because of the specific order. We are **never** calling ``set_result()``,``set_exception()``, or ``cancel()`` after ``announce_done()`` is called, which is what sets the done event.

kyleknap:
It is also worth noting that it would be a pretty big bug if it setting the result or exception is done after setting the done event because the done event is what unblocks ``result()`` so if ``set_result()`` or ``set_exception()`` is called after then ``result()`` may not returning or raising the right thing.

jamesls:
I don't know if that's a good enough guarantee.  These are "public" APIs from the perspective of s3transfer internals so what's enforcing this guarantee?  i.e it's entirely possible to write buggy code that calls this in the wrong order.  We're *assuming* that code isn't going to do this or else it's wrong.

> It is also worth noting that it would be a pretty big bug if it setting the result or exception is done after setting the done event because the done event is what unblocks result() so if set_result() or set_exception() is called after then result() may not returning or raising the right thing.

Then can we enforce this via some exception being raised?

kyleknap:
I see what you are getting at. For result/setting setting I will check to make sure the done event has not been set and raise an error if it is set to make that explicitly a rule. For what it is worth, I introduced such a bug and a bunch of the tests failed, so I think we would have caught but I will add that validation if that is good with you?

jamesls:
Yeah I think that validation would be good to add.

kyleknap:
@jamesls I added the validation like we talked about so it should be good to look at again. I had to update some of our tests and logic based on this new validation as well. The logic needed to be updated because when we tried an invalid state transition, it essentially became a no-op so it was not affected by the invalid transition attempt, but now we are throwing errors so we have to account for that in places. I put pretty big comments about those places to help explain it.

tests/unit/test_tasks.py
```diff
@@ -225,7 +226,18 @@ def test_cleanups_only_ran_once_on_exception(self):
         for done_callback in done_callbacks:
             self.transfer_coordinator.add_done_callback(done_callback)
 
-        submission_task()
+        try:
+            submission_task()
+        except InvalidStateTransitionError:
```
jamesls: I don't think we should be doing this.  The point of that exception is that seeing an invalid state transition is indicative of a bug as you mentioned in a previous comment.

Also, what's to prevent this from happening in the "real" code?  Seems like there's a chance a user could actually see an `InvalidStateTransitionError`

kyleknap: It will not be propagated to "real" code because submission tasks are ran in executors, where if they see an error, they will just set the resulting future's result to the invalid state error. The only chance a user can run into one of these is if they call ``TransferFuture.cancel()`` themselves as there is no other direct hooks to updating the status of the ``TransferCoordintor``.

kyleknap: Do you think we should be catching it [here](https://github.com/boto/s3transfer/blob/develop/s3transfer/tasks.py#L128) instead? I did not do that is a good idea because I thought it would be worse to be covering up invalid state transitions.

The real problem lies in that a single task cannot guarantee that if it tries to set the result/exception that another thread has not come in and set a result and set the done event. To do that we need to couple state setting and done event setting under a single threading primitive, which would be part of the state revamp.

Do you have any ideas?

PR #62, opened 2016-09-22
Add UnexpectedCancelledError

kyleknap:
Introduced a new type of ``CancelledError`` called ``UnexpectedCancelledError`` that differentiates between a cancellation by the user or cntrl-c and a general error hit in the context manger of the TransferManager.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -18,6 +18,7 @@ import threading
 
 from s3transfer.compat import MAXINT
 from s3transfer.exceptions import CancelledError
+from s3transfer.exceptions import UnexpectedCancelledError
 from s3transfer.utils import FunctionContainer
 from s3transfer.utils import TaskSemaphore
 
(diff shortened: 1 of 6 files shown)


s3transfer/futures.py
```diff
@@ -221,16 +222,23 @@ def result(self):
                 raise self._exception
             return self._result
 
-    def cancel(self, msg=''):
+    def cancel(self, msg='', by_user=True):
```
jamesls: Rather than `by_user`, what about providing your own cancelled error type.  That way we can extend this in the future for new cancellation types without having to add more params.

kyleknap: Makes sense

s3transfer/manager.py
```diff
@@ -482,14 +482,19 @@ def __enter__(self):
...
         # all of the inprogress futures in the shutdown.
         if exc_type:
             cancel = True
             cancel_msg = str(exc_value)
             if not cancel_msg:
                 cancel_msg = repr(exc_value)
-        self.shutdown(cancel, cancel_msg)
+            # If it was a KeyboardInterrupt, the cancellation was initiated
+            # by the user.
+            if exc_type == KeyboardInterrupt:
```
jamesls: `isinstance`

kyleknap:
@jamesls I updated based on feedback. I also changed ``UnexpectedCancelledError`` to ``FatalError`` because I did not like that name very much. Should be good to look at again.

PR #63, opened 2016-09-28
Use deferred files when opening files

kyleknap:
If the files were opened in the submission task, it would cause too many file handles to be opened. Now this gets opened in the IOExecutor which has a limit of 1000 tasks and should not be an issue because most OS's have a max of 1024 open file handles.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -34,6 +34,7 @@ from s3transfer.utils import calculate_range_parameter
 from s3transfer.utils import FunctionContainer
 from s3transfer.utils import CountCallbackInvoker
 from s3transfer.utils import StreamReaderProgress
+from s3transfer.utils import DeferredOpenFile
 from s3transfer.tasks import Task
 from s3transfer.tasks import SubmissionTask
 
@@ -190,7 +191,8 @@ class DownloadFilenameOutputManager(DownloadOutputManager):
         return f
 
     def _get_fileobj(self, filename):
-        f = self._osutil.open(filename, 'wb')
+        f = DeferredOpenFile(filename, mode='wb')
+        f.OPEN_METHOD = self._osutil.open
         # Make sure the file gets closed and we remove the temporary file
         # if anything goes wrong during the process.
         self._transfer_coordinator.add_failure_cleanup(f.close)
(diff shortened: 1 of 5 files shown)


s3transfer/download.py
```diff
@@ -171,7 +172,8 @@ def _get_temp_fileobj(self):
         return f
 
     def _get_fileobj(self, filename):
-        f = self._osutil.open(filename, 'wb')
+        f = DeferredOpenFile(filename, mode='wb')
+        f.OPEN_METHOD = self._osutil.open
```
jamesls: We're already adding a `mode` argument, why not add an `open_function` argument as well?

jamesls:
To make sure  I understand the issue, even though the request executor also has a 1000 cap on its queue size, the file is still opened when the task is done executing because it's sent off to the io executor.  As a result, it's possible to exceed more than 1024 open files right?

Otherwise change looks good to me.

s3transfer/download.py
kyleknap: Seems reasonable. I can add it. Before, I think i was only using it on tests and the upload method, but since I am doing it twice now, it seems reasonable.

kyleknap:
Yeah that is pretty much sums it up. Before, we had a bunch of open files that were not being worked on in the request executor's queue because they got opened in a submission executor and since the io executor has a queue itself there were some open file handles there too, causing us to easily spill over 1024 because both of those queues were of size 1000.

Ideally we only want to be opening the file if it is about to be worked on, not if we are preparing to work on it. The one downside I see with this is that before we would fast fail in the submission task if say we did not have permissions to the file as we opened the file there but now we will fail at the first attempt of a write to the file.

kyleknap:

s3transfer/download.py
```diff
@@ -191,8 +191,8 @@ def _get_temp_fileobj(self):
         return f
 
     def _get_fileobj(self, filename):
-        f = DeferredOpenFile(filename, mode='wb')
-        f.OPEN_METHOD = self._osutil.open
+        f = DeferredOpenFile(
+            filename, mode='wb', open_method=self._osutil.open)
```
jamesls: I would prefer this be called `open_function`, as that most accurately describes the arg (it's also how what you describe it in your docstring). Plus the default value is the `open` function.

kyleknap: Ahh, yeah I think that was because the class variable used to be called ``OPEN_METHOD`` and that threw me off a bit. I just updated it to be ``open_function``.

kyleknap:
@jamesls I updated. Should be good to look at again.

PR #64, opened 2016-09-28
Avoid io executor for nonranged downloads

kyleknap:
The io executor and the other primatives that came up with it  were unnecessary for nonranged files as there is only ever one thread working on a particular nonranged download. These were relatively pretty heavy weight so by not using the io executor and writing to the file-like object
immediately upon download improves both speed and memory.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -106,14 +106,33 @@ class DownloadOutputManager(object):
         """
         self._transfer_coordinator.submit(
             self._io_executor,
-            IOWriteTask(
-                self._transfer_coordinator,
-                main_kwargs={
-                    'fileobj': fileobj,
-                    'data': data,
-                    'offset': offset,
-                }
-            )
+            self.get_io_write_task(fileobj, data, offset)
+         )
+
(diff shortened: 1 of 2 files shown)


s3transfer/download.py
```diff
@@ -518,6 +534,21 @@ def _main(self, client, bucket, key, fileobj, extra_args, callbacks,
...
+class ImmediatelyWriteIOGetObjectTask(GetObjectTask):
+    """GetObjectTask that immediately writes to the provided file object
+
+    This is useful for downloads where it is known only one thread is
+    downloading the object so there is no reason to go through the
+    overhead of using an IO queue and executor.
+    """
+    def _handle_io(self, download_output_manager, fileobj, chunk, index):
+        task = download_output_manager.get_io_write_task(fileobj, chunk, index)
+        task()
```
jamesls: I assume there's no change in error behavior now that any IO errors for this task will be raised in the request executor instead of the io executor?

kyleknap: Yes that is exactly what will happen. The exception will be raised and exception set within one of the threads of the request executor.

PR #65, opened 2016-10-04
Introduce DownloadChunkIterator

kyleknap:
Additional logic was need to account for downloading empty objects where we still want to return the empty byte string once for the empty object. Without this logic, users cannot download an empty object by filename.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -508,8 +508,8 @@ class GetObjectTask(Task):
                     response['Body'], callbacks)
 
                 current_index = start_index
-                chunks = iter(
-                    lambda: streaming_body.read(io_chunksize), b'')
+                chunks = DownloadChunkIterator(
+                    streaming_body, io_chunksize, response)
                 for chunk in chunks:
                     # If the transfer is done because of a cancellation
                     # or error somewhere else, stop trying to submit more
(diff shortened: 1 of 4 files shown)


s3transfer/download.py
```diff
@@ -614,6 +614,38 @@ def _main(self):
...
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        chunk = self._body.read(self._chunksize)
+        self._num_reads += 1
+        if chunk:
+            return chunk
+        elif self._num_reads == 1 and not self._response.get('ContentLength'):
```
jamesls: Why do we need the `ContentLength` check?

kyleknap: Now that I think about it is not actually needed.  I originally had ugly logic where I tried to handle a lot of this logic in the ``GetObjectTask`` based off the response, but then I decided to create a new iterator for it, but since we are able to track reads we do not need to check ``ContentLength`` anymore. I will update the class.

kyleknap:
@jamesls I updated this PR to refactor the class a bit. Should be good to look at again.

s3transfer/download.py
jamesls: That seems fine.   The other alternative if you don't want to have to check every single time if `_num_reads` is not `1` is to just make `__iter__` a generator, check the content length header once, and if it's greater than 0 then iterate over the body.

```
def __iter__(self):
    if self.headers['Content-Length'] == 0:
        yield b''
    else:
        for chunk in iter(self._body, lambda: ....):
            yield chunk
```

kyleknap: True that would work as well.

PR #67, opened 2016-10-05
Remove use of unique_id() for handler registration

kyleknap:
The ``unique_id()`` was not needed because the same enable/disable callback
handler can be used across ``TransferManagers``. When using ``unique_id()``, it
would cause a memory leak when the same client was used for multiple
``TransferManagers`` as the handler would always get registered on
instantiation.

Fixes https://github.com/boto/boto3/issues/819

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/manager.py
+++ b/s3transfer/manager.py
@@ -470,12 +470,12 @@ class TransferManager(object):
     def _register_handlers(self):
         # Register handlers to enable/disable callbacks on uploads.
         event_name = 'request-created.s3'
-        enable_id = 's3upload-callback-enable'
-        disable_id = 's3upload-callback-disable'
         self._client.meta.events.register_first(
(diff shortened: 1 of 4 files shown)


s3transfer/manager.py
```diff
@@ -471,8 +470,8 @@ def _get_submission_task_main_kwargs(
     def _register_handlers(self):
         # Register handlers to enable/disable callbacks on uploads.
         event_name = 'request-created.s3'
-        enable_id = unique_id('s3upload-callback-enable')
-        disable_id = unique_id('s3upload-callback-disable')
+        enable_id = 's3upload-callback-enable'
```
jamesls: Do we need to store these as local vars?  They're only used once now

kyleknap:

PR #68, opened 2016-10-06
Support downloading to FIFOs

JordonPhillips:
Previously FIFOs had been lumped in with other "special files",
but they act differently than normal files. They are non-seekable,
in-memory files which typically require an active reader and
writer.

This adds a new output manager for FIFOs which treats them as
in-memory non-seekable file objects whose source is a deferred file
handle with buffering disabled.

Resolves aws/aws-cli#2217

cc @kyleknap @jamesls

### DIFF
--- a/s3transfer/utils.py
+++ b/s3transfer/utils.py
@@ -19,9 +19,11 @@ import stat
 import string
 import logging
 import threading
+import io
 from collections import defaultdict
 
 from s3transfer.compat import rename_file
+from s3transfer.compat import seekable
 
 
 MAX_PARTS = 10000
@@ -256,7 +258,7 @@ class OSUtils(object):
         # Block special device
         if stat.S_ISBLK(mode):
             return True
-        # FIFO.
+        # Named pipe / FIFO
         if stat.S_ISFIFO(mode):
             return True
         # Socket.
@@ -269,7 +271,7 @@ class DeferredOpenFile(object):
     def __init__(self, filename, start_byte=0, mode='rb', open_function=open):
         """A class that defers the opening of a file till needed
 
-        This is useful for deffering opening of a file till it is needed
+        This is useful for deferring opening of a file till it is needed
         in a separate thread, as there is a limit of how many open files
         there can be in a single thread for most operating systems. The
         file gets opened in the following methods: ``read()``, ``seek()``,
@@ -296,7 +298,8 @@ class DeferredOpenFile(object):
     def _open_if_needed(self):
         if self._fileobj is None:
             self._fileobj = self._open_function(self._filename, self._mode)
-            self._fileobj.seek(self._start_byte)
+            if self._start_byte != 0:
+                self._fileobj.seek(self._start_byte)
 
     @property
     def name(self):
--- a/tests/unit/test_download.py
+++ b/tests/unit/test_download.py
@@ -14,6 +14,7 @@ import copy
 import os
 import shutil
 import tempfile
+import mock
 
 from tests import BaseTaskTest
 from tests import BaseSubmissionTaskTest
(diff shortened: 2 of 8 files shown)


kyleknap:
@jamesls @JordonPhillips What was the direction did we want to go with this PR? Wasn't it that for any special file we will treat it as nonseekable and not use a temporary file and not just for FIFO?

jamesls:
I think that makes sense.  I believe we have the same problem right now for domain sockets and it'd be nice to handle that case as well.

JordonPhillips:
@jamesls sockets are another pile of worms, you need to open them with `socket.connect`

jamesls:
@JordonPhillips fair, the case I was thinking of actually requires you to stream data through stdout/stdin already, so you'd never specify the path to a domain socket directly in the CLI/s3transfer.

s3transfer/utils.py
```diff
@@ -296,7 +298,8 @@ def __init__(self, filename, start_byte=0, mode='rb', open_function=open):
     def _open_if_needed(self):
         if self._fileobj is None:
             self._fileobj = self._open_function(self._filename, self._mode)
-            self._fileobj.seek(self._start_byte)
+            if self.seekable:
```
kyleknap: Ahh I see why you added this now. I wonder if we can get away with something simpler like add a parameter to the ``__init__`` to ensure we do not seek to start or maybe even reusing ``start_byte`` to have the meaning of ``None`` be to not seek anywhere to start?

tests/unit/test_download.py
```diff
@@ -175,38 +177,33 @@ def test_get_file_io_write_task(self):
...
         with self.download_output_manager.get_fileobj_for_io_writes(
                 self.future) as f:
             # Ensure it is a file like object returned
             self.assertTrue(hasattr(f, 'read'))
-            self.assertTrue(hasattr(f, 'seek'))
             # Make sure the name of the file returned is the same as the
             # final filename as we should not be writing to a temporary file.
             self.assertEqual(f.name, self.filename)
 
     def test_get_final_io_task(self):
```
kyleknap: Might be good to add some of the tests from the non-seekable test cases to make sure it behaves as what we would expect as well, specifically the io task submission happens in order of offset.

s3transfer/utils.py
JordonPhillips: That's what I had initially, but the feedback was to use seekable instead

JordonPhillips:
The number of commits was getting out of control, so I squashed.

tests/unit/test_utils.py
```diff
@@ -269,30 +271,33 @@ def test_is_special_file_for_non_existant_file(self):
...
-        super(TestDefferedOpenFile, self).tearDown()
+        super(TestDeferredOpenFile, self).tearDown()
 
-    def counting_open_function(self, filename, mode):
-        self.open_called_count += 1
+    def recording_open_function(self, filename, mode):
+        self.open_call_args.append((filename, mode))
         return open(filename, mode)
 
+    def recording_open_non_nonseekable(self, filename, mode):
```
kyleknap: This is not used anymore.

s3transfer/utils.py
```diff
@@ -296,7 +298,8 @@ def __init__(self, filename, start_byte=0, mode='rb', open_function=open):
     def _open_if_needed(self):
         if self._fileobj is None:
             self._fileobj = self._open_function(self._filename, self._mode)
-            self._fileobj.seek(self._start_byte)
+            if self._start_byte != 0:
```
kyleknap: I think we should add a test to make sure we do not seek anymore when start byte is 0.

PR #69, opened 2016-10-17
Set exception on future

JordonPhillips:
This allows setting a future to the failure state by setting
the exception. This can be useful in callbacks.

cc @kyleknap @jamesls

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -77,6 +77,9 @@ class TransferFuture(object):
         """Cancels the request associated with the TransferFuture"""
         self._coordinator.cancel()
 
+    def set_exception(self, exception):
+        self._coordinator.set_exception(exception)
+
 
 class TransferMeta(object):
     """Holds metadata about the TransferFuture"""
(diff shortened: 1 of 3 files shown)


s3transfer/futures.py
```diff
@@ -77,6 +77,9 @@ def cancel(self):
         """Cancels the request associated with the TransferFuture"""
         self._coordinator.cancel()
 
+    def set_exception(self, exception):
```
kyleknap: We can add a docstring to this?

JordonPhillips:
So what I've done is changed `set_exception` in `TransferCoordinator` to noop if an exception has already been set rather than just checking for `done` since our use case is to set failure from success. `set_exception` on `TransferFuture` then raises an error if you try to use it before the transfer is done.

JordonPhillips:
Capturing some offline discussion: We do want to be able to re-set the exception after the transfer is done from the `TransferFuture` but not have that be the default behavior. Update adds an `override` flag.

PR #70, opened 2016-10-19
Add non-threaded executor implementation

kyleknap:
This pull request adds support for supplying a custom executor implementation. Specifically it allows you to use a non-threaded version where all of the logic lives in the main thread. To do this I exposed an `executor_cls` parameter on `TransferManager.__init__()`.

I also needed to update our testing scripts to test the functionality of this implementation by reusing the functional and integration tests. Here are some important decisions I made in doing this where feedback would be good:
- The non-threaded version of the tests can be ran by setting the environment variable `USE_SERIAL_IMPLEMENTATION`.
- In `ci/` scripts I decided not to run any of the plugins like xunit or coverage because I do not want that run to clobber any of the xml files that we get back when running the normal implementation.
- I skipped tests where it is was relying on the main thread not being blocked by the transfer happening in the main thread such as FIFO tests and the Cntrl-C tests. In the end, I was really trying to avoid modifying any of these existing tests, but for the Cntrl-C tests there is not really a great way to inject the Cntrl-C into the process without spinning up an entirely new process. So I was not really sure what to do about that. Suggestions would be great. I also tried signalling an alarm with a KeyboardInterrupt, but that is a bit dangerous as if it got raised at an inappropriate location it would exit out of the tests completely. For what it is worth, I did manual testing and I was able to Cntrl-C very quickly.

In the end, the hope is to use this executor implementation to fix: https://github.com/boto/boto3/issues/814. I did a quick test with the script provided in the issue and it now successfully downloads the object (i.e. no more hanging) so that is good.

cc @jamesls @JordonPhillips

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -14,9 +14,11 @@ from concurrent import futures
 from collections import namedtuple
 import copy
 import logging
+import sys
 import threading
 
 from s3transfer.compat import MAXINT
+from s3transfer.compat import six
 from s3transfer.exceptions import CancelledError, TransferNotDoneError
 from s3transfer.utils import FunctionContainer
 from s3transfer.utils import TaskSemaphore
@@ -465,6 +467,76 @@ class ExecutorFuture(object):
 
     def done(self):
         return self._future.done()
+
+
+class NonThreadedExecutor(object):
+    """A drop-in replacement non-threaded version of ThreadPoolExecutor"""
+    def __init__(self, max_workers=None):
+        pass
+
+    def submit(self, fn, *args, **kwargs):
+        future = NonThreadedExecutorFuture()
+        try:
+            result = fn(*args, **kwargs)
+            future.set_result(result)
+        except BaseException:
+            e, tb = sys.exc_info()[1:]
+            logger.debug(
+                'Setting exception for %s to %s with traceback %s',
+                future, e, tb
+            )
+            future.set_exception_info(e, tb)
+        return future
+
+    def shutdown(self, wait=True):
+        pass
+
+
+class NonThreadedExecutorFuture(object):
+    """The Future returned from NonThreadedExecutor
+
+    Note that this future is **not** thread-safe as it is being used
+    from the context of a non-threaded environment.
+    """
+    def __init__(self):
+        self._result = None
+        self._exception = None
+        self._traceback = None
+        self._done = False
+        self._done_callbacks = []
+
+    def set_result(self, result):
+        self._result = result
+        self._set_done()
+
+    def set_exception_info(self, exception, traceback):
+        self._exception = exception
+        self._traceback = traceback
+        self._set_done()
+
+    def result(self, timeout=None):
+        if self._exception:
+            six.reraise(
+                type(self._exception), self._exception, self._traceback)
+        return self._result
+
+    def _set_done(self):
+        self._done = True
+        for done_callback in self._done_callbacks:
+            self._invoke_done_callback(done_callback)
+        self._done_callbacks = []
+
+    def _invoke_done_callback(self, done_callback):
+        return done_callback(self)
+
+    def done(self):
+        return self._done
+
+    def add_done_callback(self, fn):
+        if self._done:
+            self._invoke_done_callback(fn)
+        else:
+            self._done_callbacks.append(fn)
 
 
 TaskTag = namedtuple('TaskTag', ['name'])
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -370,7 +370,7 @@ class BoundedExecutor(object):
     EXECUTOR_CLS = futures.ThreadPoolExecutor
 
     def __init__(self, max_size, max_num_threads, tag_semaphores=None,
-                 underlying_executor_cls=None):
+                 executor_cls=None):
         """An executor implentation that has a maximum queued up tasks
 
         The executor will block if the number of tasks that have been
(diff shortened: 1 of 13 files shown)


s3transfer/futures.py
```diff
@@ -367,7 +369,8 @@ def _run_callback(self, callback):
 class BoundedExecutor(object):
     EXECUTOR_CLS = futures.ThreadPoolExecutor
 
-    def __init__(self, max_size, max_num_threads, tag_semaphores=None):
+    def __init__(self, max_size, max_num_threads, tag_semaphores=None,
+                 underlying_executor_cls=None):
```
jamesls: I don't think the `underlying` adds anything, this is being passed as a init param so it suggests that it's going to be used by this class.  I'd prefer just `executor_cls` like it's done in `manager.py`.

s3transfer/tasks.py
```diff
@@ -253,7 +253,7 @@ def _main(self, transfer_future, **kwargs):
             # Call the submit method to start submitting tasks to execute the
             # transfer.
             self._submit(transfer_future=transfer_future, **kwargs)
-        except Exception as e:
+        except BaseException as e:
```
jamesls: Why was this changed?  I didn't see this mentioned in the PR description.

scripts/ci/run-tests
```diff
@@ -12,9 +12,13 @@ REPO_ROOT = _dname(_dname(_dname(os.path.abspath(__file__))))
...
-    return check_call(command, shell=True)
+def run(command, env=None):
+    return check_call(command, shell=True, env=env)
 
 
 run('nosetests --with-coverage --cover-erase --cover-package s3transfer '
     '--with-xunit --cover-xml -v unit/ functional/')
+
+# Run the serial implementation of s3transfer
+os.environ['USE_SERIAL_IMPLEMENTATION'] = 'True'
```
jamesls: serial implementation is somewhat vague, implementation of what?  Can we call it `USE_SERIAL_EXECUTOR`?

tests/__init__.py
```diff
@@ -34,13 +34,23 @@
...
 from s3transfer.futures import BoundedExecutor
+from s3transfer.futures import NonThreadedExecutor
 from s3transfer.subscribers import BaseSubscriber
 from s3transfer.utils import OSUtils
 from s3transfer.utils import CallArgs
 from s3transfer.utils import TaskSemaphore
 from s3transfer.utils import SlidingWindowSemaphore
 
 
+def setup_package():
```
jamesls: While not strictly necessary, I still think it's good practice to put back any global state you mutated.  I think we should have a `teardown_package()` as well.

tests/functional/test_manager.py
```diff
@@ -125,3 +139,12 @@ def test_enable_disable_callbacks_only_ever_registered_once(self):
...
             'once')
+
+    def test_use_custom_executor_implementation(self):
+        transfer_manager = TransferManager(
+            self.client, executor_cls=BaseExecutor)
+        # The BaseExecutor class has no methods implemented so it should
+        # immediately raise a NotImplementedError once something is
+        # submitted to it
+        with self.assertRaises(NotImplementedError):
+            transfer_manager.delete('bucket', 'key')
```
jamesls: I don't think this is necessarily a strict contract we have.  I.e if we decide to move implementation into the base class, that's perfectly reasonable, and yet this test will fail.

I think it would better to just pass in a mock for `executor_cls` here.

tests/unit/test_futures.py
```diff
@@ -530,6 +546,14 @@ def test_shutdown_no_wait(self):
         # not done, which it should not be because it it slow.
         self.assertFalse(future.done())
 
+    def test_replace_underlying_executor(self):
+        executor = BoundedExecutor(10, 1, {}, BaseExecutor)
+        # The BaseExecutor class has no methods implemented so it should
+        # immediately raise a NotImplementedError once something is
+        # submitted to it
+        with self.assertRaises(NotImplementedError):
```
jamesls: Same thing, either use a mock or create a stub implementation.

s3transfer/futures.py
kyleknap: Makes sense. I will update that.

scripts/ci/run-tests
kyleknap: Yeah that's fine.

tests/__init__.py
kyleknap: Makes sense

tests/functional/test_manager.py
kyleknap: I think I am going to create a stub implementation to cover both of the ``NotImplementedError`` comments.

s3transfer/tasks.py
kyleknap: It is changed specifically for the serial implementation. The ``SubmissionTask`` acts as the outer most layer of the main thread when the serial implementation is used and so if a ``KeyboardInterrupt`` is received during the transfer process, done will not properly get announced, causing the process to hang if using the context manager of the TransferManager, because ``Exception`` does not catch the ``KeyboardInterrupt``.

I will add a comment to make that more clear.

kyleknap:

s3transfer/futures.py
```diff
@@ -385,9 +388,16 @@ def __init__(self, max_size, max_num_threads, tag_semaphores=None):
...
             number of tasks the executor is processing at a time.
+
+        :type executor_cls: BaseExecutor
+        :param underlying_executor_cls: The executor class that
+            get bounded by this executor. If None is provided, the
+            concurrent.futures.ThreadPoolExecutor class is used.
         """
         self._max_num_threads = max_num_threads
-        self._executor = self.EXECUTOR_CLS(max_workers=self._max_num_threads)
+        if not executor_cls:
```
jamesls: if executor_cls is None

s3transfer/tasks.py
```diff
@@ -253,11 +253,17 @@ def _main(self, transfer_future, **kwargs):
...
             # If there was an exception rasied during the submission of task
             # there is a chance that the final task that signals if a transfer
             # is done and too run the cleanup may never have been submitted in
             # the first place so we need to account accordingly.
+            #
+            # Note that BaseException is caught, instead of Exception, because
+            # for some implmentations of executors, specifically the serial
+            # implementation, the SubmissionTask is directly exposed to
+            # KeyboardInterupts and so need to cleanup and signal done
```
jamesls: "and so needs to cleanup"

kyleknap:
Made the changes. Added a changelog entry and test pass. Merging.

PR #88, opened 2017-06-12
Implement seek whence param for zero length files

kamalmostafa:
Bug: https://github.com/aws/aws-cli/issues/2403
Bug-Ubuntu: https://bugs.launchpad.net/bugs/1696800

As described in the bug reports referenced above, trying to copy a
zero-length file to S3 fails (with some version combinations of python3,
awscli, s3transfer, and requests) like this:

    $ aws s3 cp emptyfile s3://somebucket
    upload failed: ./emptyfile to s3://somebucket/emptyfile seek() takes 2 positional arguments but 3 were given

... due to the incomplete implementation of seek() -- missing the optional
'whence' argument -- in s3transfer's ReadFileChunk method.  This patch
adds the whence argument to all of s3transfer's seek implementations.

### DIFF
--- a/s3transfer/utils.py
+++ b/s3transfer/utils.py
@@ -313,9 +313,9 @@ class DeferredOpenFile(object):
         self._open_if_needed()
         self._fileobj.write(data)
 
-    def seek(self, where):
+    def seek(self, where, whence=0):
         self._open_if_needed()
-        self._fileobj.seek(where)
+        self._fileobj.seek(where, whence)
 
     def tell(self):
         if self._fileobj is None:
@@ -439,8 +439,8 @@ class ReadFileChunk(object):
     def disable_callback(self):
         self._callbacks_enabled = False
 
-    def seek(self, where):
-        self._fileobj.seek(self._start_byte + where)
+    def seek(self, where, whence=0):
+        self._fileobj.seek(self._start_byte + where, whence)
         if self._callbacks is not None and self._callbacks_enabled:
             # To also rewind the callback() for an accurate progress report
             invoke_progress_callbacks(
--- a/s3transfer/__init__.py
+++ b/s3transfer/__init__.py
@@ -276,8 +276,8 @@ class ReadFileChunk(object):
     def disable_callback(self):
         self._callback_enabled = False
 
-    def seek(self, where):
-        self._fileobj.seek(self._start_byte + where)
+    def seek(self, where, whence=0):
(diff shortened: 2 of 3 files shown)


s3transfer/utils.py
```diff
@@ -439,8 +439,8 @@ def enable_callback(self):
     def disable_callback(self):
         self._callbacks_enabled = False
 
-    def seek(self, where):
-        self._fileobj.seek(self._start_byte + where)
+    def seek(self, where, whence=0):
```
kyleknap: So I think this implementation of seek will be a bit more tricky than the other ones. Mainly because the `ReadFileChunk` class has an additional abstraction that controls where you start and limits how far you can read into the file. Using the `self._start_byte` for the beginning of the file and `self._size` for the end of the file, this class ensures you cannot read before the beginning or after the end. So I think that passing in `whence` of non-zero breaks this abstraction. To fix it, I think it would be reasonable to:

1) Calculate the `where` value that is passed into `self._fileobj.seek()` based on the value of `whence` passed in along with the `self._start_byte` and `self._size`. So for `0`, where would be `self._start_byte + where`. For `1`, `self._amount_read + where`. For `2`, `self._size + where`.

2) Always pass `0` for `whence` into `self._fileobj.seek()`

3) Use the same calculated value in the progress callback and when setting `self._amount_read`

Let me know if that makes sense.

kamalmostafa: Thanks Kyle.  Yes, that does make sense.  Can I leave it to you to implement that, given your familiarity with the codebase and my lack thereof?  (I'll be happy to verify that your implementation still fixes the zero-length files problem).

queeno:
+1

bschuon:
+1

irishman2020:
+1

s3transfer/utils.py
kamalmostafa: Ping (kyleknap)!  This problem is still unresolved.  Is your proposed fix ready to merge (or post here as a patch for verification testing)?  Thanks!

dbnicholson: I took the liberty of handling the comments above. My PR is in #102.

bschuon:
*Bump*

stephgosling:
_another bump_

d33tah:
bump

CesarPantoja:
Still an issue

nateprewitt:
Resolving this in favor of the merged PR #102 that addressed the feedback here. Thanks everyone!

PR #89, opened 2017-07-05
Subclass bytes to support the Buffer API

dstufft:


jamesls:
Not sure what to make of the travis failures, any ideas?

dstufft:
I'm not sure, I'm poking at it but it appears the event handler that is being registered by TransferManager isn't getting called at all.

tipabu:
FWIW, I think the failure had to do with `CallbackEnablingBody` always being a False-y value, meaning that when we'd prepare the request, we'd pass a `None` body to requests. See https://github.com/boto/botocore/blob/@{2017-07-05}/botocore/vendored/requests/models.py#L455 -- since `data` is False-y, `body` stays `None`.

Even with modern botocore (which has its own request-preparer, no longer vendors requests, etc.) the `body == b''` check in `AWSRequestPreparer._prepare_body` (see https://github.com/boto/botocore/blob/1.12.26/botocore/awsrequest.py#L411-L412, for example) would cause the same behavior.

Regardless, pretty sure this got addressed by #91, when `CallbackEnablingBody` started subclassing `RawIOBase` -- I think this PR can be closed.

PR #91, opened 2017-08-21
Don't mask unicode exceptions

### LINKED ISSUE #85: Unicode error masking upload errors
Source: boto/boto3#1009

Basically [this](https://github.com/boto/s3transfer/blob/develop/s3transfer/manager.py#L499) is masking actual errors being raised if they have unicode characters.

JordonPhillips:
Fixes #85

### DIFF
--- a/.changes/next-release/bugfix-TransferManager-84566.json
+++ b/.changes/next-release/bugfix-TransferManager-84566.json
@@ -1,0 +1,5 @@
+{
+  "category": "TransferManager", 
+  "type": "bugfix", 
+  "description": "Properly handle unicode exceptions in the context manager. Fixes `#85 <https://github.com/boto/boto3/issues/85>`__"
+}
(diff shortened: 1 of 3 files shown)


joguSD:
PR looks fine to me. Travis is failing but it seems a bad merge into the `develop` branch is the root cause.

kyleknap:
Yeah I think the tests needed to be updated after the sigv4 by default switch. I think it would be best to get the tests updated first so it is stable before getting this PR merged.

JordonPhillips:
Yep I got it fixed. Just had to disable payload signing.

Integralist:
Hello, 

I was googling a specific internal client error and I stumbled across this PR (which is now merged and looks to fix the error). I see this has gone into `boto:develop` branch, but when is this likely to end up in `master` (what's the release process like for this lib?) and so subsequently I'm guessing once in master it'll end up exposed in the boto3 library?

Thanks.

PR #95, opened 2017-09-28
Accept ContentMD5 in ExtraArgs for s3 file uploads

m-vdb:
Hi there,

I've just migrated from `boto` to `boto3` and with `boto` we were able to pass a `content-md5` header that would ensure data integrity server-side (following [this guide](https://aws.amazon.com/premiumsupport/knowledge-center/data-integrity-s3/)). It appears that it is no longer possible with `boto3`. After 30 minutes debugging and navigating between the 3 projects (`boto3`, `botocore` and `s3transfer`) I finally manager to pinpoint the issue.

`ContentMD5` is missing from the accepted arguments when doing a file upload. Also, I tried to use the `Metadata` argument to provide it, but it didn't work because each header passed in `Metadata` is prefixed using `x-amz-meta-`, which doesn't work at all for the desired `content-md5` header.

So this PR is an attempt to fix this. I've tested that locally and it seems to work properly. LMK if I need to add more things. (I'll fix tests if there are failing of course)

### DIFF
--- a/s3transfer/manager.py
+++ b/s3transfer/manager.py
@@ -155,13 +155,12 @@ class TransferManager(object):
         'RequestPayer',
     ]
 
-    ALLOWED_UPLOAD_ARGS = [
+    BASE_ALLOWED_UPLOAD_ARGS = [
         'ACL',
         'CacheControl',
         'ContentDisposition',
         'ContentEncoding',
         'ContentLanguage',
-        'ContentMD5',
         'ContentType',
         'Expires',
         'GrantFullControl',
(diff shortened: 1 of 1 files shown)


tholam:
I was also looking for information about this and it seems like botocore tries to add the content-md5 header to every operation already.  https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L948-L949  

I might've just missed it but I don't think I saw it documented anywhere so it might be useful to create a PR for boto3 to add a mention of this in the docs.

PR #97, opened 2017-10-21
Returns client response from upload_file

### LINKED ISSUE #82: upload_file return value
Both `PutObject` and `CompleteMultipartUpload` respond with data that includes the `VersionId` and `ETag`. [1] [2]

It would be really useful if S3Transfer could return this response, or some part of the response.

 - [1] http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.complete_multipart_upload
 - [2] http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_object

aweidner:
Fixes #82 

The only thing I'm not sure of here is the testing.  I'm not a fan of having multiple assert statements in the same test cases but I don't want to duplicate the existing test code unnecessarily.  Thoughts?

### DIFF
--- a/s3transfer/__init__.py
+++ b/s3transfer/__init__.py
@@ -391,7 +391,7 @@ class MultipartUploader(object):
             raise S3UploadFailedError(
                 "Failed to upload %s to %s: %s" % (
                     filename, '/'.join([bucket, key]), e))
-        self._client.complete_multipart_upload(
+        return self._client.complete_multipart_upload(
             Bucket=bucket, Key=key, UploadId=upload_id,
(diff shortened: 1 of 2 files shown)


HurSungYun:
I really want this. Thanks for fix it

ricbartm:
@aweidner I'm interested on this to be reviewed by upstream too. Would you mind to tag the issue with the label "needs-review" to bring better visibility over it?

aweidner:
@ricbartm Seems only those with write access to the repository are allowed to apply a label to a pull request.  I'd love to get this reviewed if possible though, it has been open for a while now.

PR #101, opened 2017-11-28
Add max_bandwidth option

kyleknap:
Allows user to configure the max bandwidth consumption for the streaming of data for uploads and downloads. The value is set in terms of bytes per second. I would recommend trying it out with the CLI PR that I am planning to send up soon.

### DIFF
--- a/s3transfer/bandwidth.py
+++ b/s3transfer/bandwidth.py
@@ -1,0 +1,409 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import time
+import threading
+
+
+class RequestExceededException(Exception):
+    def __init__(self, requested_amt, retry_time):
+        """Error when requested amount exceeds what is allowed
+
+        The request that raised this error should be retried after waiting
+        the time specified by ``retry_time``.
+
+        :type requested_amt: int
+        :param requested_amt: The originally requested byte amount
+
+        :type retry_time: float
+        :param retry_time: The length in time to wait to retry for the
+            requested amount
+        """
+        self.requested_amt = requested_amt
+        self.retry_time = retry_time
+        msg = (
+            'Request amount %s exceeded the amount available. Retry in %s' % (
+                requested_amt, retry_time)
+        )
+        super(RequestExceededException, self).__init__(msg)
+
+
+class RequestToken(object):
+    """A token to pass as an identifier when consuming from the LeakyBucket"""
+    pass
+
+
+class TimeUtils(object):
+    def time(self):
+        """Get the current time back
+
+        :rtype: float
+        :returns: The current time in seconds
+        """
+        return time.time()
+
+    def sleep(self, value):
+        """Sleep for a designated time
+
+        :type value: float
+        :param value: The time to sleep for in seconds
+        """
+        return time.sleep(value)
+
+
+class BandwidthLimiter(object):
+    def __init__(self, token_bucket, time_utils=None):
+        """Limits bandwidth for shared S3 transfers
+
+        :type leaky_bucket: LeakyBucket
+        :param token_bucket: The leaky bucket to use limit bandwidth
+
+        :type time_utils: TimeUtils
+        :param time_utils: Time utility to use for interacting with time.
+        """
+        self._token_bucket = token_bucket
+        self._time_utils = time_utils
+        if time_utils is None:
+            self._time_utils = TimeUtils()
+
+    def get_bandwith_limited_stream(self, fileobj, transfer_coordinator,
+                                    enabled=True):
+        """Wraps a fileobj in a bandwidth limited stream wrapper
+
+        :type fileobj: file-like obj
+        :param fileobj: The file-like obj to wrap
+
+        :type transfer_coordinator: s3transfer.futures.TransferCoordinator
+        param transfer_coordinator: The coordinator for the general transfer
+            that the wrapped stream is a part of
+
+        :type enabled: boolean
+        :param enabled: Whether bandwidth limiting should be enabled to start
+        """
+        stream = BandwidthLimitedStream(
+            fileobj, self._token_bucket, transfer_coordinator,
+            self._time_utils)
+        if not enabled:
+            stream.disable_bandwidth_limiting()
+        return stream
+
+
+class BandwidthLimitedStream(object):
+    def __init__(self, fileobj, leaky_bucket, transfer_coordinator,
+                 time_utils=None, bytes_threshold=256 * 1024):
+        """Limits bandwidth for reads on a wrapped stream
+
+        :type fileobj: file-like object
+        :param fileobj: The file like object to wrap
+
+        :type token_bucket: LeakyBucket
+        :param token_bucket: The leaky bucket to use to throttle reads on
+            the stream
+
+        :type transfer_coordinator: s3transfer.futures.TransferCoordinator
+        param transfer_coordinator: The coordinator for the general transfer
+            that the wrapped stream is a part of
+
+        :type time_utils: TimeUtils
+        :param time_utils: The time utility to use for interacting with time
+        """
+        self._fileobj = fileobj
+        self._leaky_bucket = leaky_bucket
+        self._transfer_coordinator = transfer_coordinator
+        self._time_utils = time_utils
+        if time_utils is None:
+            self._time_utils = TimeUtils()
+        self._bandwidth_limiting_enabled = True
+        self._request_token = RequestToken()
+        self._bytes_seen = 0
+        self._bytes_threshold = bytes_threshold
+
+    def enable_bandwidth_limiting(self):
+        """Enable bandwidth limiting on reads to the stream"""
+        self._bandwidth_limiting_enabled = True
+
+    def disable_bandwidth_limiting(self):
+        """Disable bandwidth limiting on reads to the stream"""
+        self._bandwidth_limiting_enabled = False
+
+    def read(self, amount):
+        """Read a specified amount
+
+        Reads will only be throttled if bandwidth limiting is enabled.
+        """
+        if not self._bandwidth_limiting_enabled:
+            return self._fileobj.read(amount)
+
+        # We do not want to be calling consume on every read as the read
+        # amounts can be small causing the lock of the leaky bucket to
+        # introduce noticeable overhead. So instead we keep track of
+        # how many bytes we have seen and only call consume once we pass a
+        # certain threshold.
+        self._bytes_seen += amount
+        if self._bytes_seen < self._bytes_threshold:
+            return self._fileobj.read(amount)
+
+        self._consume_through_leaky_bucket()
+        return self._fileobj.read(amount)
+
+    def _consume_through_leaky_bucket(self):
+        # NOTE: If the read amonut on the stream are high, it will result
+        # in large bursty behavior as there is not an interface for partial
+        # reads. However given the read's on this abstraction are at most 256KB
+        # (via downloads), it reduces the burstiness to be small KB bursts at
+        # worst.
+        while not self._transfer_coordinator.exception:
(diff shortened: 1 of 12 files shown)


jamesls:
Looking at this now, but would you mind adding a more detailed commit message?  This is a pretty big feature and I think it's worth more of an explanation.

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+        # amounts can be small causing the lock of the leaky bucket to
+        # introduce noticeable overhead. So instead we keep track of
+        # how many bytes we seen and only call consume once we pass a
+        # certain threshold.
+        self._bytes_seen += amount
+        if self._bytes_seen < self._bytes_threshold:
+            return self._fileobj.read(amount)
+
+        self._consume_through_leaky_bucket()
+        return self._fileobj.read(amount)
```
dstufft: This is going to result in very bursty reads I think in cases where someone is passing a large number. I don't think there's a great way around this besides allowing partial reads which I think would be a change in the contract. However it might make sense to at least document that large numbers passed to amount might reduce the effectiveness of bandwidth limiting.

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+        )
+        super(RequestExceededException, self).__init__(msg)
+
+
+class RequestToken(object):
+    """A token to pass as an identifier when consuming from the LeakyBucket"""
+    pass
+
+
+class TimeUtils(object):
```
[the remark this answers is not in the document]
kyleknap: It is so you inject your own implementations. We do this a lot in `s3transfer` with `OSUtils` and this pattern is done a lot with `chalice`. If you didn't, you would have to patch a bunch of stuff to affect the behavior of `time`.

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+            self._last_time = time_at_consumption
+            self._current_rate = 0.0
+            return
+        self._current_rate = self._calculate_ema_rate(amt, time_at_consumption)
+        self._last_time = time_at_consumption
+
+    def _calculate_rate(self, amt, time_at_consumption):
+        return amt/(time_at_consumption - self._last_time)
+
+    def _calculate_ema_rate(self, amt, time_at_consumption):
```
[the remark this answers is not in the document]
kyleknap: exponential moving average. I can see if I can rename it or add a comment.

s3transfer/bandwidth.py
kyleknap: Yep that is true if they were passing any a large number. The good news is they really do not have control over the value passed into the `read()` for this abstraction. For downloads, the internals of `s3transfer` will always read 256KB when streaming data down, and for uploads `httplib` uses 8KB (which we don't even have control over). So given that the burstiness will be pretty small if at all.

s3transfer/bandwidth.py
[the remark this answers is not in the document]
joguSD: What I mean is we don't need a wrapper class to do this. `time_module=time` would also work as our restraint is that the thing passed has a `time` and `sleep` callable. The only benefit I see is that there is an example of what interface the `time_module` needs to meet.

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import time
+import threading
+
+
+class RequestExceededException(Exception):
+    def __init__(self, requested_amt, retry_time):
+        """Error
```
JordonPhillips: This isn't very helpful

s3transfer/bandwidth.py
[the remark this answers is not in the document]
JordonPhillips: If we ever need anything more complicated in the future, this will allow us to have it. OSUtils has things like that which aren't direct mappings.

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+
+        :type leaky_bucket: LeakyBucket
+        :param token_bucket: The leaky bucket to use limit bandwidth
+
+        :type time_utils: TimeUtils
+        :param time_utils: Time utility to use for interacting with time.
+        """
+        self._token_bucket = token_bucket
+        self._time_utils = time_utils
+        if not time_utils:
```
JordonPhillips: `time_utils is not None`

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+            that the wrapped stream is a part of
+
+        :type time_utils: TimeUtils
+        :param time_utils: The time utility to use for interacting with time
+        """
+        self._fileobj = fileobj
+        self._leaky_bucket = leaky_bucket
+        self._transfer_coordinator = transfer_coordinator
+        self._time_utils = time_utils
+        if not time_utils:
```
JordonPhillips: `time_utils is not None`

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+            self._time_utils = TimeUtils()
+        self._enabled = True
+        self._request_token = RequestToken()
+        self._bytes_seen = 0
+        self._bytes_threshold = bytes_threshold
+
+    def enable(self):
```
JordonPhillips: How about something like `enable_bandwidth_limiting` (and similar for disabled)? This makes it seem like you're enabling / disabling the object itself rather than the limiting feature.

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+
+        Reads will only be throttled if bandwidth limiting is enabled.
+        """
+        if not self._enabled:
+            return self._fileobj.read(amount)
+
+        # We do not want to be calling consume on every read as the read
+        # amounts can be small causing the lock of the leaky bucket to
+        # introduce noticeable overhead. So instead we keep track of
+        # how many bytes we seen and only call consume once we pass a
```
JordonPhillips: we have seen

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+                self._schedule_during_max_bandwidth_usage(
+                    amt, request_token, time_now)
+            else:
+                return self._release_tokens(amt, time_now)
+
+    def _will_exceed_max_bandwidth_capacity(self, amt, time_now):
+        projected_rate = self._rate_tracker.get_projected_rate(amt, time_now)
+        return projected_rate > self._max_rate
+
+    def _consume_for_scheduled_consumption(self, amt, request_token, time_now):
```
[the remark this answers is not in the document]
JordonPhillips: +1

s3transfer/bandwidth.py
jamesls: I think more worrisome for me is the fact that this seems to rely on multiple threads to smooth over the burstiness.  The less threads there are the more bursting you get.  This also appears to throw off the bandwidth calculations.  I'm not quite sure why this happens, but it's really pronounced if you set the bandwidth to something high and the max concurrent requests to 1.  In one test, I had 100MB+ bandwidth available, and with one thread and `max_bandwidth` to something really high (order of GB/s) I was only getting about 15MB/s.

Any insight on what's going on here?  I haven't had a whole lot of time to dig into this.

kyleknap:
@JordonPhillips I think I got all of your comments

@joguSD I addressed all of your comments. Did not change the `TimeUtils` though. I feel like I would prefer to keep it as is mainly because of Jordon's point that it ensures we are not coupled to the `time` module and implementers do not have to implement everything under the `time` module to have a compliant implementation.

@dstufft I put a comment about what you were talking about for large transfers. Still looking at @jamesls brought up. Let me know if there was something specific you were thinking?

@jamesls I am still working at your stuff, but yeah I will add a better commit. I'm probably going to squash all of these and make a better commit before it gets merged in.

kyleknap:
Also as to debug logging. I do not have any set currently. To track retries and consumption amount I had an abstraction that I could query before, but I yanked that from the original review. How does these logs sound?:

* Logging for when bandwidth limiting gets set in the transfer manager
* Logging the request token and the retry time when requests get schedueled for retries
* Logging when a scheduled requests is completed and releases that amt.

s3transfer/manager.py
```diff
@@ -481,11 +503,11 @@ def _register_handlers(self):
...
         event_name = 'request-created.s3'
         self._client.meta.events.register_first(
-            event_name, disable_upload_callbacks,
-            unique_id='s3upload-callback-disable')
+            event_name, signal_not_transferring,
+            unique_id='s3upload-not-transferring')
         self._client.meta.events.register_last(
-            event_name, enable_upload_callbacks,
-            unique_id='s3upload-callback-enable')
+            event_name, signal_transferring,
```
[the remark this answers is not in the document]
kyleknap: I am not a big fan of it as well, but essentially what it suppose to encompass is to indicate that the stream is being read because it is being transferred. This is needed because the stream for uploads can be read for the following reasons:

* It is being read and uploaded to S3
* It is being read for content MD5 
* It is being read in memory to calculate the SHA256 for a sigv4 request.

The problem is that when you have functionality that is specific to the file stream being read from for the specific purpose of uploading to s3 (i.e. transfer callbacks and transfer bandwidth limiting), it is ambiguous on why the read is happening. We handle the MD5 point to start because these are disabled to start. However, the SHA256 portion is tricky because the signing happens during the request creation and `request-created` is the last event we have before sending the request over the wire. To solve it, we straddle the handler that does the sha256 calculation by disabling to start and enable it back at the very end. Hence the meaning of `signal_not_transferring()` comes to be "even though the stream is being read from, it is not actually being sent to S3 and the meaning of `signal_transferring()` means "if the stream is being read from it is because it is being read from in transferring the data".

I would be up for better names if that is confusing, but I am not sure if there is a better mechanism to use without exposing something new in botocore. Not sure if you have any thoughts?

s3transfer/bandwidth.py
```diff
@@ -0,0 +1,395 @@
...
+            except RequestExceededException as e:
+                self._time_utils.sleep(e.retry_time)
+        else:
+            raise self._transfer_coordinator.exception
+
+    def signal_transferring(self):
+        """Signal that data being read is being transferred to S3"""
+        self.enable()
+        if hasattr(self._fileobj, 'signal_transferring'):
+            self._fileobj.signal_transferring()
```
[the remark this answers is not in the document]
kyleknap: I explained the signal transferring portion in the comment.

As to the `hasattr` check that is because of a whole other issue. The issue boils down to we need one shared event handler with no shared state that can indicate to a stream that the files is being read for the specific reason of being transferred to S3. We need one shared handler because one handler per request really bogs down the system. We need no share state because the handler could be invoked by separate threads.

So to achieve this, the current strategy is to check for a method on the stream and call it in the handlers to enable/disable transfer reading functionality. This is what we had in place for callbacks. However if the stream is wrapped multiple times and two of these wrappers require this enabling/disabling functionality, before only the top level wrapper would have its method called and the wrapper underneath would not have its method called. So instead the top level wrapper needs to be able to signal to the wrapper underneath that it was signaled to enable/disable whatever functionality it needs to enable/disable and so it will need to as well. So I tried to generalize all of this with the `signal_transferring()` and `signal_not_transferring()` methods. I'm up for suggestions on how to improve this.

s3transfer/bandwidth.py
kyleknap: I tried looking into it. I was running on a machine where with default configuration's of 10 threads I was getting 100MB+, but I was never able to get a single thread (with bandwidth enabled or not enabled) to ever reach that. It was seeing speeds of ~25 MB with bandwidth turned off but turning on to something high resulted in the same speeds. I saw this for both uploads and downloads. I want to see it on your environment though.

As to the bandwidth calculations, those are just bursty in general because we aggregate the amounts we report back so if only one thread is working it will report in increments of that threshold. When there is more threads reporting it smoothes the progress updates more because there are more threads forcing the the progress to update. So that may explain that? If the threshold is low and you have one thread working, I can see the progress even more bursty because there is only thread reporting and it has to do sleeps.

s3transfer/manager.py
```diff
@@ -247,6 +254,16 @@ def __init__(self, client, config=None, osutil=None, executor_cls=None):
...
             executor_cls=executor_cls
         )
+
+        # The component responsible for limiting bandwidth usage if it
+        # is configured.
+        self._bandwidth_limiter = None
+        if self._config.max_bandwidth is not None:
+            logging.debug(
+                'Setting max_bandwidth to %s', self._config.max_bandwidth)
+            token_bucket = LeakyBucket(self._config.max_bandwidth)
```
jamesls: s/token_bucket/leaky_bucket/g

PR #103, opened 2018-02-14
Plumb request payer to complete multipart

kyleknap:
I noticed that this was not plumbed in when I was working on the `--request-payer` feature for the CLI.

### DIFF
--- a/s3transfer/utils.py
+++ b/s3transfer/utils.py
@@ -126,22 +126,22 @@ def invoke_progress_callbacks(callbacks, bytes_transferred):
             callback(bytes_transferred=bytes_transferred)
 
 
-def get_filtered_extra_args(extra_args, whitelisted_args):
-    """Gets a filtered dictionary of arguments
+def get_filtered_dict(original_dict, whitelisted_keys):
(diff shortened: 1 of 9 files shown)


s3transfer/utils.py
```diff
@@ -126,6 +126,24 @@ def invoke_progress_callbacks(callbacks, bytes_transferred):
             callback(bytes_transferred=bytes_transferred)
 
 
+def get_filtered_extra_args(extra_args, whitelisted_args):
```
jamesls: Can you rename this to something more general?  The name makes sense when it's an internal method on the various Task subclasses, but if it's going to be a utility function I think it should be named without any mention of "extra args".

kyleknap: Yeah I was thinking about that too. I am going to rename it to `get_filtered_dict`.

kyleknap:
@jamesls I updated based on feedback. Let me know if you have any feedback on the update. Otherwise, I will probably merge this soon so we can start getting clean runs of the CLI `--request-payer` PR.

PR #105, opened 2018-05-15
Fix TransferManager.cancel passing bad parameters to _shutdown

danielkza:
Passing the `cancel` boolean twice tries to pass a string as the exception class, which eventually crashes when creating an instance of it is attempted.


guoshimin2:
@nateprewitt Sorry for tagging you directly. This is a blatant bug. Could you merge the fix please?

PR #96, opened 2017-10-12
Add Tagging and TaggingDirective to upload parameters

DennisTT:
Issue: https://github.com/boto/s3transfer/issues/94
Used in: https://github.com/aws/aws-cli/pull/2890

PR #107, opened 2018-06-06
Added object-level-tagging to ExtraArgs

m6a-UdS:
S3 now supports Object-Level tagging using the Tagging parameter using the `S3.Client.put_object`, but using `upload_fileobj ` does not because the underlying `s3transfer` does not allow it. This PR allows the parameter to be passed and enables upload-time object-level tagging.


m6a-UdS:
Oops, duplicate of: https://github.com/boto/s3transfer/pull/96

PR #112, opened 2018-10-20
Fix corrupted S3 downloads when file object in append mode

hhamalai:
The DownloadSeekableOutputManager with threading causes file corruption when the output file is opened in append mode. 

The file object is still seekable, but only for reads, the writes will be written into the end of the file.

This is also the root cause for #1446, and fixes the issue.


s3transfer/download.py
```diff
@@ -24,7 +24,7 @@
     ReadTimeoutError
 
 from s3transfer.compat import SOCKET_ERROR
-from s3transfer.compat import seekable
+from s3transfer.compat import seekable, writes_are_seekable
```
swt2c: Just a fly-by comment here, but unless I'm missing something, you don't actually call this `writes_are_seekable()` function anywhere, so how does this fix the problem?

hhamalai: Excellent that someone is awake, added the missing call from my staging 👍

PR #121, opened 2019-01-23
Implement future interface for ProcessPoolDownloader

kyleknap:
There are still some improvements that I would like to add. I put these as TODO comments in the code. I will follow up on these with PR's. If you see anything else that may be an issue let me know. I will either fix it or add a TODO.

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -68,7 +68,7 @@ class BaseTransferMeta(object):
     @property
     def user_context(self):
         """A dictionary that requesters can store data in"""
-        raise NotImplementedError('transfer_id')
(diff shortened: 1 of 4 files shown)


s3transfer/futures.py
```diff
@@ -27,7 +27,51 @@
...
+
+    @property
+    def transfer_id(self):
+        """The unique id of the transfer"""
+        raise NotImplementedError('transfer_id')
+
+    @property
+    def user_context(self):
+        """A dictionary that requesters can store data in"""
+        raise NotImplementedError('transfer_id')
```
stealthycoin: `user_context`

PR #131, opened 2019-02-08
Fix unbound local error

### LINKED ISSUE #120: download.py::GetObjectTask::_main : set current_index earlier in the function
I'm getting an error saying

```python
...
File "/usr/local/lib/python3.6/dist-packages/s3transfer/download.py", line 554, in _main
callbacks, start_index - current_index)
UnboundLocalError: local variable 'current_index' referenced before assignment
```

meaning there's an exception in the `except` clause of the `try-except` of the function.

Line 554 being https://github.com/boto/s3transfer/blob/f506f396f054d35288bcf86d5289f8749a366ccd/s3transfer/download.py#L554

And indeed, if an error is raised before `current_index = start_index` in line 530
https://github.com/boto/s3transfer/blob/develop/s3transfer/download.py#L530

then this will happen. Perhaps line 530 can be moved upwards so `current_index` gets a value? Or initialize it with `None` so that I can tell whether the exception was raised before/after line 530?

Thanks!

jamesls:
The first commit cleans up a few linting issues, then https://github.com/boto/s3transfer/commit/70586dcd10af2de7426f78b8e64a1cae3c887973 fixes #120.

### DIFF
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -11,8 +11,6 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import logging
-import os
-import socket
 import threading
 import heapq
 
@@ -100,7 +98,7 @@ class DownloadOutputManager(object):
         self._transfer_coordinator.submit(
             self._io_executor,
             self.get_io_write_task(fileobj, data, offset)
-         )
+        )
 
     def get_io_write_task(self, fileobj, data, offset):
         """Get an IO write task for the requested set of data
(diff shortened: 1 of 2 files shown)


jamesls:
@kyleknap I'm not sure what's going on with travis.  There was an issue on py3 with the tests but I've fixed it.  However, it's still running the tests against the old commit so I'll try a rebase/force push, but this PR is ready to look at.

jamesls:
Adding the rest of the team.  Build is passing now.

PR #144, opened 2019-10-21
Use correct parameter list to _shutdown()

JasonDeArte:
This update fixes a param call typo, to a member function 2 lines below it.

Previous to this change, the "cancel_msg" param would get a boolean instead of a prefered string.


tschoonj:
Can someone merge this in please? It is a trivial fix for an obvious bug.

tschoonj:
See also #117 and #105

PR #151, opened 2019-12-19
add additional transfer config parameters for processpool

rajeshwar-nu:
implemented a TODO item and added config parameters for io chunsize and max attempts to the `ProcessTransferConfig`


rajeshwar-nu:
FYI @stealthycoin @kyleknap can I get reviews on this?

PR #153, opened 2020-01-08
Drop support for Python 2.6 and 3.3

joguSD:


setup.py
```diff
@@ -38,7 +38,7 @@ def get_version():
     include_package_data=True,
     install_requires=requires,
     extras_require={
-        ':python_version=="2.6" or python_version=="2.7"': [
```
kyleknap: We need to update this in the `setup.cfg` as well.

PR #158, opened 2020-01-16
fixing temp filename length to be less than or equal to 255

swetashre:
fixes issue https://github.com/boto/botocore/issues/1916

### DIFF
--- a/s3transfer/utils.py
+++ b/s3transfer/utils.py
@@ -239,6 +239,8 @@ class CountCallbackInvoker(object):
 
 
 class OSUtils(object):
+    _MAX_FILENAME_LEN = 255
+
     def get_file_size(self, filename):
         return os.path.getsize(filename)
 
@@ -300,7 +302,11 @@ class OSUtils(object):
         return False
 
     def get_temp_filename(self, filename):
-        return filename + os.extsep + random_file_extension()
+        suffix = os.extsep + random_file_extension()
+        path = os.path.dirname(filename)
+        name = os.path.basename(filename)
+        temp_filename = name[:self._MAX_FILENAME_LEN - len(suffix)] + suffix 
+        return os.path.join(path, temp_filename)
 
     def allocate(self, filename, size):
         try:
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -305,6 +305,16 @@ class TestOSUtils(BaseUtilsTest):
             )
         )
 
+    def test_get_temp_filename_len_255(self):
+        filename = 'a'*255
+        temp_filename = OSUtils().get_temp_filename(filename)
+        self.assertLessEqual(len(temp_filename), 255)
+    
+    def test_get_temp_filename_len_gt_255(self):
+        filename = 'a'*280
+        temp_filename = OSUtils().get_temp_filename(filename)
+        self.assertLessEqual(len(temp_filename), 255)
+
     def test_allocate(self):
         truncate_size = 1
         OSUtils().allocate(self.filename, truncate_size)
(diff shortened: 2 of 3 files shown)


tests/unit/test_utils.py
```diff
@@ -297,10 +297,11 @@ def test_is_special_file_for_non_existant_file(self):
         self.assertFalse(OSUtils().is_special_file(non_existant_filename))
 
     def test_get_temp_filename(self):
```
joguSD: This test shouldn't need to be updated as the filename is well under the max of 255 (and shouldn't change). Let's add a new test where the filename is over 255 and assert that it's truncated to the known length there.

s3transfer/utils.py
```diff
@@ -300,7 +300,12 @@ def is_special_file(cls, filename):
         return False
 
     def get_temp_filename(self, filename):
-        return filename + os.extsep + random_file_extension()
+        max_filename_len = 255
+        sep = os.extsep
+        ext = random_file_extension()
+        path,name = os.path.dirname(filename), os.path.basename(filename)
+        temp_filename = name[:max_filename_len-len(sep)-len(ext)] + sep + ext 
+        return os.path.join(path,temp_filename)
```
joguSD: Let's add a space after the comma here.

s3transfer/utils.py
```diff
@@ -300,7 +300,12 @@ def is_special_file(cls, filename):
         return False
 
     def get_temp_filename(self, filename):
-        return filename + os.extsep + random_file_extension()
+        max_filename_len = 255
+        sep = os.extsep
+        ext = random_file_extension()
+        path,name = os.path.dirname(filename), os.path.basename(filename)
+        temp_filename = name[:max_filename_len-len(sep)-len(ext)] + sep + ext
```
joguSD: Let's add spaces around the operators like so `max_filename_len - len(sep) - len(ext)`.

If the line starts to get too long it might make sense to break it up and store some of these calculations in a variable and then use them in the array range.

s3transfer/utils.py
```diff
@@ -300,7 +302,12 @@ def is_special_file(cls, filename):
         return False
 
     def get_temp_filename(self, filename):
-        return filename + os.extsep + random_file_extension()
+        sep = os.extsep
+        ext = random_file_extension()
+        path = os.path.dirname(filename)
+        name = os.path.basename(filename)
+        temp_filename = name[:self._MAX_FILENAME_LEN - len(sep) - len(ext)] + sep + ext
```
joguSD: This line is getting a little long, let's split it up so it's under 80 characters. Considering we always use `sep` and `ext` together it might make sense to just combine them into a `suffix` like:
```
suffix = os.extsep + random_file_extension()
...
temp_filename = name[:self._MAX_FILENAME_LEN - len(suffix)] + suffix
```

tests/unit/test_utils.py
```diff
@@ -304,6 +304,13 @@ def test_get_temp_filename(self):
                 OSUtils().get_temp_filename(filename)
             )
         )
+    def test_get_temp_filename_len_255(self):
```
joguSD: Let's add a newline above this. Functions inside of a class should be separated by a single new line.

tests/unit/test_utils.py
```diff
@@ -304,6 +304,13 @@ def test_get_temp_filename(self):
...
+        filename = 'a'*255
+        self.assertTrue(len(OSUtils().get_temp_filename(filename))<=255)
```
joguSD: Let's use the `assertLessEqual` method instead of `<=`.  It might also make sense to break out the `get_temp_filename` call. Something like:
```python
temp_filename = OSUtils().get_temp_filename(filename)
self.assertLessEqual(len(temp_filename), 255)
```

joguSD:
This is looking good, just had a couple more small comments on formatting / readability.

PR #173, opened 2021-01-05
Update to add server side encryption context into allowed list

annguojia:
Add x-amz-server-side-encryption-context into allowed upload args


stealthycoin (approved):
Looks good.

nateprewitt (approved):
:shipit:

PR #179, opened 2021-03-08
futures.TransferCoordinator: fix cancellation deadlock

tschoonj:
This patches fixes a deadlock whenever `TransferManager._shutdown` is
called with cancel==True


tschoonj:
@kdaily could you give this one a look? And #144?

Thanks in advance!!

PR #185, opened 2021-03-24
Fix trove classifiers for s3transfer

nateprewitt:
Update the Trove classifiers in s3transfer to accurately reflect the versions of Python we currently test and support.


zdutta (approved):
👌

PR #186, opened 2021-03-30
Crt develop

TingDaoK:


tests/functional/test_crt.py
```diff
@@ -0,0 +1,229 @@
+import mock
+import unittest
+try:
+    import awscrt.s3
+except ImportError:
+    raise unittest.SkipTest("skip CRT test")
```
kyleknap, line 6: I did a a bit more research on what the skipping options are out there. Looks like there is not great options with regards to skipping all tests in a module and avoid import errors. For example even though CI passed for this one, the skip message is not great: https://github.com/boto/s3transfer/runs/2223526536#step:5:57

Instead, I think we do the following:
1. Remove all of these `ImportError` try/excepts and then `SkipTest` raising
2. Guard all imports of `awscrt` and `s3transfer.crt` with an `if HAS_CRT` check at the module level
3. Update all usage of classes from `awscrt` and `s3transfer.crt` to include the full path of the module. This is to reduce the number of imports that we have to do at the module level and makes it a little easier to follow if the class is coming from a crt module that may or may not be imported.

I already started sketching out what this would look. Here is a sample of what I was thinking for the `unit/test_crt.py` test file:
```diff
diff --git a/tests/functional/test_crt.py b/tests/functional/test_crt.py
index 6cbdae3..957ebf3 100644
--- a/tests/functional/test_crt.py
+++ b/tests/functional/test_crt.py
@@ -31,6 +31,7 @@ class submitThread(threading.Thread):
         self._futures.append(self._transfer_manager.download(*self._callargs))
 
 
+# Move these two serialization classes to be defined in their single test case
 class SerializationException(Exception):
     pass
 
diff --git a/tests/unit/test_crt.py b/tests/unit/test_crt.py
index 1ed964c..1b4813a 100644
--- a/tests/unit/test_crt.py
+++ b/tests/unit/test_crt.py
@@ -1,17 +1,14 @@
 
 import unittest
-try:
-    import awscrt.s3
-except ImportError:
-    raise unittest.SkipTest("skip CRT test")
 
 from botocore.session import Session
-from s3transfer.crt import BotocoreCRTRequestSerializer
-from s3transfer.crt import CRTTransferCoordinator, CRTTransferFuture, CRTTransferMeta
 from s3transfer.utils import CallArgs
 
 from tests import FileCreator
-from tests import requires_crt
+from tests import requires_crt, HAS_CRT
+
+if HAS_CRT:
+    import s3transfer.crt
 
 
 @requires_crt
@@ -20,7 +17,8 @@ class TestBotocoreCRTRequestSerializer(unittest.TestCase):
         self.region = 'us-west-2'
         self.session = Session()
         self.session.set_config_variable('region', self.region)
-        self.request_serializer = BotocoreCRTRequestSerializer(self.session)
+        self.request_serializer = s3transfer.crt.BotocoreCRTRequestSerializer(
+            self.session)
         self.bucket = "test_bucket"
         self.key = "test_key"
         self.files = FileCreator()
@@ -35,9 +33,9 @@ class TestBotocoreCRTRequestSerializer(unittest.TestCase):
         callargs = CallArgs(
             bucket=self.bucket, key=self.key, fileobj=self.filename,
             extra_args={}, subscribers=[])
-        coordinator = CRTTransferCoordinator()
-        future = CRTTransferFuture(
-            CRTTransferMeta(call_args=callargs),
+        coordinator = s3transfer.crt.CRTTransferCoordinator()
+        future = s3transfer.crt.CRTTransferFuture(
+            s3transfer.crt.CRTTransferMeta(call_args=callargs),
             coordinator)
         crt_request = self.request_serializer.serialize_http_request(
             "put_object", future)
@@ -50,9 +48,9 @@ class TestBotocoreCRTRequestSerializer(unittest.TestCase):
         callargs = CallArgs(
             bucket=self.bucket, key=self.key, fileobj=self.filename,
             extra_args={}, subscribers=[])
-        coordinator = CRTTransferCoordinator()
-        future = CRTTransferFuture(
-            CRTTransferMeta(call_args=callargs),
+        coordinator = s3transfer.crt.CRTTransferCoordinator()
+        future = s3transfer.crt.CRTTransferFuture(
+            s3transfer.crt.CRTTransferMeta(call_args=callargs),
             coordinator)
         crt_request = self.request_serializer.serialize_http_request(
             "get_object", future)
@@ -65,9 +63,9 @@ class TestBotocoreCRTRequestSerializer(unittest.TestCase):
         callargs = CallArgs(
             bucket=self.bucket, key=self.key,
             extra_args={}, subscribers=[])
-        coordinator = CRTTransferCoordinator()
-        future = CRTTransferFuture(
-            CRTTransferMeta(call_args=callargs),
+        coordinator = s3transfer.crt.CRTTransferCoordinator()
+        future = s3transfer.crt.CRTTransferFuture(
+            s3transfer.crt.CRTTransferMeta(call_args=callargs),
             coordinator)
         crt_request = self.request_serializer.serialize_http_request(
             "delete_object", future)

```

tests/functional/test_crt.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class SerializationException(Exception):
+    pass
+
+
+if HAS_CRT:
+    class ExceptionRaisingSerializer(s3transfer.crt.BaseCRTRequestSerializer):
+        def serialize_http_request(self, transfer_type, future):
+            raise SerializationException()
```
kyleknap, line 37: Since we only use this class once I was thinking we just move it into the test case. So something like this to reduce the number of `if HAS_CRT` checks:
```python
    def test_serializer_error_handling(self):
        class SerializationException(Exception):
             pass

       class ExceptionRaisingSerializer(s3transfer.crt.BaseCRTRequestSerializer):
           def serialize_http_request(self, transfer_type, future):
               raise SerializationException()


        not_impl_serializer = ExceptionRaisingSerializer()
        transfer_manager = s3transfer.crt.CRTTransferManager(
            crt_s3_client=self.s3_crt_client,
            crt_request_serializer=not_impl_serializer)
        future = transfer_manager.upload(
            self.filename, self.bucket, self.key, {}, [])

        with self.assertRaises(SerializationException):
            future.result()
```

kyleknap (commented):
Looks fine. We should also squash all of the commits related to getting the conditional tests to pass into a single commit.

PR #187, opened 2021-04-13
new adapter

TingDaoK:


s3transfer/crt.py
```diff
@@ -22,6 +22,27 @@
 logger = logging.getLogger(__name__)
 
 
+class CRTCredentialProviderAdapter:
+    def __init__(self, botocore_credential_provider):
+        self._botocore_credential_provider = botocore_credential_provider
+        self._loaded_credentials = None
+        self._lock = threading.Lock()
+
+    def __call__(self):
```
kyleknap, line 31: Let's add two unit tests for this class:

1. One case that is just the happy case where we make sure the correct methods are called and the correct `AwsCredentials` object is returned back
2. Another case to make sure we only call `load_credentials()` once on multiple `__call__()` invocations.

kyleknap (commented):
Looks good. Let's just get some unit tests added and we should be set.

kyleknap (approved):
Thanks!

PR #189, opened 2021-04-14
Fix setup.cfg botocore[crt] specifier

nateprewitt:
Quick update to fix a typo in the original test branch. Botocore 1.20.29 was the first version to release support for the CRT extra.


kyleknap (approved):
:ship:

PR #191, opened 2021-04-14
Add support for optional AWS Common Runtime (CRT) bindings

kyleknap:
AWS CRT provides a C-based S3 transfer client that can improve transfer throughput for uploads and downloads. You will now be able to install `s3transfer` with `crt` as an optional dependency:
```
pip install s3transfer[crt]
```
And then use transfer manager abstractions from the `s3transfer.crt` module in order to leverage the performance from the CRT S3 transfer client while using a similar interface as the `s3transfer.manager.TransferManager` class.

### DIFF
--- a/.changes/next-release/feature-crt-30324.json
+++ b/.changes/next-release/feature-crt-30324.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "``crt``",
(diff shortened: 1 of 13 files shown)


nateprewitt (approved):
Looks good :shipit:

PR #194, opened 2021-04-19
Update setup.py

Assimilationstheorie:
Please be consistent. With everything. :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,8 @@ import sys
 
 from setuptools import setup, find_packages
 
-
 ROOT = os.path.dirname(__file__)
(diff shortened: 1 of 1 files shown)


nateprewitt (changes requested):
Hi @Assimilationstheorie, thanks for the PR! We may want to invert the current proposed change.

If you'd like to add a single newline after line 14, that will bring us in-line with standard Python style which is to have two lines between top-level elements.

nateprewitt:
Going to resolve this as it hasn't had an update in a few months. Please feel free to reopen if you're able to address the current feedback.

PR #218, opened 2021-10-14
Extend linting tools

nateprewitt:
This is a continuation of #205 to add some more substantial framing for how we handle code structure in the boto projects. This PR is currently a draft for comment and may be broken up into smaller parts for easier review. The current proposal adds 3 new tools to ensure we're cleaning up old code and keeping what we have currently consistent. This also helps expedite PR review by removing the need for iterations on minor style fixes.

### Proposed additions:
* **isort** - This will order our imports and ensure consistent layout. This avoids duplicate imports by keeping everything in the same spot and makes it easier to find what's already available.
* **pyupgrade** - This will automatically "upgrade" existing code to use the feature set configured for the lowest supported version (in this case Python 3.6). This is a more significant change that most version bumps because it's cleaning up all of our Python 2.6, 2.7 and 3.0-3.5 code.
* **black** -  This is the last one and maybe most contentious. We're adopting black with two caveats, we're not going to enforce quote normalization for now, and we're going to keep line length at 79 for the time being.

### DIFF
--- a/scripts/performance/benchmark
+++ b/scripts/performance/benchmark
@@ -96,9 +96,19 @@ def run_benchmark(pid, output_file, data_interval):
             recv_rate = recv_delta / dt
 
             # Save all of the data into a CSV file.
-            f.write(','.join(str(val) for val in [
-                current_time, memory_used, cpu_percent, sent_rate,
-                recv_rate]) + '\n')
+            f.write(
+                ','.join(
+                    str(val)
+                    for val in [
+                        current_time,
+                        memory_used,
+                        cpu_percent,
+                        sent_rate,
+                        recv_rate,
+                    ]
+                )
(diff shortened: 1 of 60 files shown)


scripts/performance/benchmark
```diff
@@ -97,9 +96,19 @@ def run_benchmark(pid, output_file, data_interval):
             recv_rate = recv_delta / dt
 
             # Save all of the data into a CSV file.
-            f.write(','.join(str(val) for val in [
-                current_time, memory_used, cpu_percent, sent_rate,
-                recv_rate]) + '\n')
+            f.write(
```
joguSD, line 99: This might be something we just refactor since it was pretty crazy to start.

joguSD (approved):

PR #244, opened 2022-06-26
Returned ETag for uploaded entities

adamsc64:
**Summary of Changes**
- If 'ETag' is present in service response, return it inside a metadata dictionary for both `tasks.CompleteMultipartUploadTask` and `upload.PutObjectTask`.
- This change will allow integrations which rely on these call sites to access the ETag.

**Background**
- There's a feature request over at `boto3` to return ETag information for several call sites, i.e.: https://github.com/boto/boto3/issues/2861. In order to achieve this, I believe we need `s3transfer` to return the ETag information at the call sites changed in this Pull Request.

**Design decisions**
- I was unsure how to model the entity which would be returning the ETag information. I chose to follow [the pattern](https://github.com/boto/s3transfer/blob/develop/s3transfer/upload.py#L795-L802) I found inside `upload.UploadPartTask`, i.e. a pattern already in the codebase. This pattern returns the ETag inside an outer `metadata` dictionary.
- Let me know if I should use another strategy, or if this general approach to the problem is problematic for some reason.

### DIFF
--- a/s3transfer/tasks.py
+++ b/s3transfer/tasks.py
@@ -377,11 +377,17 @@ class CompleteMultipartUploadTask(Task):
             ``UploadPartTask.main()``.
         :param extra_args:  A dictionary of any extra arguments that may be
             used in completing the multipart transfer.
+        :returns: A dictionary representing metadata information about the
+            upload, such as the ETag for the uploaded entity.
         """
-        client.complete_multipart_upload(
+        response = client.complete_multipart_upload(
             Bucket=bucket,
             Key=key,
             UploadId=upload_id,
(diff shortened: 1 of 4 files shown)


adamsc64:
@kyleknap, I don't know what you think of these changes. Can I get a review? Also open to suggestions if you think someone else is better for the moment. Thanks so much.

adamsc64:
Hi @kyleknap. New contributor to boto and s3transfer here! Wondering what is the correct way to officially request a review of this by the s3transfer team. Thanks!

cfxegbert:
It could be nice to return the `versionId` (optionally None) field also.

Issue #86, opened 2017-05-16
Ensure all pieces come from the same object while do multi-part downloading.


shileiyu:
I have done a quick search against the latest code on develop branch, the IFMatch prediction wasn't used in multi-part downloading task.
if someone patches or replaces the file during downloading(large file or slow network), the result may be end up as a mix-in of multiple revisions, especially when chunk size is small.

jamesls:
Just to clarify, the idea is to take the ETag value we get during the `head_object` call and pass it through all the `get_object()` calls.  If we get a 412 error then we mark the object as a failed (or retry again from the beginning).  Is that right?

rickleeming:
This is still a problem. Using boto3 or s3transfer to download an object may result in a corrupted download if an
upload to the same object key occurs around the same time. If the object is larger than the multipart threshold (default
8 MiB), the downloaded file may contain a mixture of the old and new versions of the object.

I can replicate it reliably if I patch s3transfer to add a sleep to the thread that downloads the first part of the
object, then upload a new version of the object while the download is running.

Code to download the object:
```
$ cat example.py
import boto3

s3 = boto3.client("s3")
s3.download_file("REDACTED", "example", "download")
```
I'm using the latest commit from the `develop` branch:
```
$ git branch
* develop

$ git log | head -n 4
commit 2a72c160ebc47a90af4e8449ce6209a6412cece9
Merge: afa90552 9ea9296c
Author: aws-sdk-python-automation <[github-aws-sdk-python-automation@amazon.com](mailto:github-aws-sdk-python-automation@amazon.com)>
Date:   Mon May 12 18:12:33 2025 +0000
```

Patch to s3transfer:
```
$ git diff
diff --git a/s3transfer/download.py b/s3transfer/download.py
index 4d431a4..536eb7a 100644
--- a/s3transfer/download.py
+++ b/s3transfer/download.py
@@ -560,6 +560,9 @@ class GetObjectTask(Task):
         :param bandwidth_limiter: The bandwidth limiter to use when throttling
             the downloading of data in streams.
         """
+        if start_index == 0:
+            import time
+            time.sleep(20)
         last_exception = None
         for i in range(max_attempts):
             try:
```

Create two files to upload:
```
$ yes a | dd iflag=fullblock bs=1M count=16 of=upload_a
$ yes b | dd iflag=fullblock bs=1M count=16 of=upload_b
```

Run the test:
```
$ python3 -m venv .venv
$ . .venv/bin/activate
[...]

PR #254, opened 2022-11-09
Specify VersionId when possible on multipart downloads

mspdt22:
This addresses #86 for versioned buckets

### DIFF
--- a/s3transfer/__init__.py
+++ b/s3transfer/__init__.py
@@ -534,6 +534,11 @@ class MultipartDownloader:
     def download_file(
         self, bucket, key, filename, object_size, extra_args, callback=None
     ):
+        self.download_versioned_file(bucket, key, filename, object_size, None, extra_args, callback=callback)
+
+    def download_versioned_file(
+        self, bucket, key, filename, object_size, object_version, extra_args, callback=None
+    ):
         with self._executor_cls(max_workers=2) as controller:
             # 1 thread for the future that manages the uploading of files
             # 1 thread for the future that manages IO writes.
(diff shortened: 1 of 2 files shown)


mspdt22:
Thank you for taking a look!

> How is this different than using VersionId in the extra_args parameter?

To clarify, are you proposing that the `s3transfer` library plumbs the VersionId from the HEAD request it already makes to get object size? Or that the `s3transfer` users themselves are responsible for making the HEAD request outside the `s3transfer` library and passing it to `download_file()`?

> One useful feature is grabbing the version id at the beginning (if available) and using that version id throughout the multipart download.

Yes, agreed. This PR is adding the feature.

cfxegbert:
> Thank you for taking a look!

FYI, I do not have any permissions to do any sort of approval on this repo.

> > How is this different than using VersionId in the extra_args parameter?
> 
> To clarify, are you proposing that the `s3transfer` library plumbs the VersionId from the HEAD request it already makes to get object size? Or that the `s3transfer` users themselves are responsible for making the HEAD request outside the `s3transfer` library and passing it to `download_file()`?

Looking at the code the real bug is that extra_args are not passed along in MultipartDownload.  If that is fixed, then `VersionId` can be passed as an extra_arg.

> > One useful feature is grabbing the version id at the beginning (if available) and using that version id throughout the multipart download.
> 
> Yes, agreed. This PR is adding the feature.

Looking at the documentation for VersionId this would be a breaking change because `GetObjectVersion` permission is required.  A non-breaking change would be to extract the `ETag` from the head_object and use it with the `IfMatch` header.

PR #255, opened 2022-11-10
fix args sent to shutdown

jbarz1:
fixes following error when submitting a cancellation message.

```
s3transfer/futures.py", line 279, in cancel
self._exception = exc_type(msg)
TypeError: 'str' object is not callable
```


cfxegbert:
See pull request #253 for integration test as well as the fix

PR #273, opened 2023-08-15
Resolve Python 3.12 warnings

nateprewitt:
This PR will clear up our Python 3.12 builds by updating our testing dependencies to avoid most of the new deprecation warnings.

We'll also turn on CRT testing since the 3.11 wheels are forward compatible.


jonathan343 (approved):
:shipit:

PR #274, opened 2023-09-22
Add user provided SSE-C arguments to CompleteMultipartUpload call

nateprewitt:
This PR will forward SSE-C arguments provided by the user to the final CompleteMultipartUpload call to handle cases where a policy has been defined which requires these for all PutObject calls.

This is to resolve a recently published workflow by S3 which changed the requiredness of these arguments due to `PutObject` being used under the hood for CompleteMultipartUpload.

### DIFF
--- a/s3transfer/upload.py
+++ b/s3transfer/upload.py
@@ -521,7 +521,13 @@ class UploadSubmissionTask(SubmissionTask):
         'ExpectedBucketOwner',
     ]
 
-    COMPLETE_MULTIPART_ARGS = ['RequestPayer', 'ExpectedBucketOwner']
+    COMPLETE_MULTIPART_ARGS = [
+        'SSECustomerKey',
+        'SSECustomerAlgorithm',
+        'SSECustomerKeyMD5',
+        'RequestPayer',
+        'ExpectedBucketOwner',
+    ]
 
     def _get_upload_input_manager_cls(self, transfer_future):
         """Retrieves a class for managing input for an upload based on file type
--- a/s3transfer/copies.py
+++ b/s3transfer/copies.py
@@ -69,7 +69,13 @@ class CopySubmissionTask(SubmissionTask):
         'TaggingDirective',
     ]
 
-    COMPLETE_MULTIPART_ARGS = ['RequestPayer', 'ExpectedBucketOwner']
+    COMPLETE_MULTIPART_ARGS = [
+        'SSECustomerKey',
+        'SSECustomerAlgorithm',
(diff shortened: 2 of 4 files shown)


s3transfer/__init__.py
```diff
@@ -380,6 +380,12 @@ class MultipartUploader:
         'RequestPayer',
     ]
 
+    COMPLETE_MULTIPART_ARGS = [
```
kyleknap, line 383: Oh we do not need to update this module. This is module is no longer used in Boto3 and CLI. It was the old s3 logic that boto3 used before migrating over to the transfer manager interface. I think we should just make sure the transfer manager code path is updated to support this and not touch this module or it's tests.

s3transfer/upload.py
```diff
@@ -521,7 +521,12 @@ class UploadSubmissionTask(SubmissionTask):
         'ExpectedBucketOwner',
     ]
 
-    COMPLETE_MULTIPART_ARGS = ['RequestPayer', 'ExpectedBucketOwner']
+    COMPLETE_MULTIPART_ARGS = [
+        'RequestPayer',
+        'ExpectedBucketOwner',
+        'SSECustomerKey',
+        'SSECustomerAlgorithm',
```
kyleknap, line 528: Do we need to proxy over the `SSECustomerKeyMD5` to the complete MPU as well? I noticed that it was in the upload parts list but not the complete mpu list.

kyleknap (commented):
Looks good. I just had a couple of comments

s3transfer/__init__.py
nateprewitt: Hmm, all of our upload tests that check this are currently written against this implementation 😬

s3transfer/upload.py
nateprewitt: There was back and forth on this. They said only the Key/Algo for now. I've added what we know we need and we can always extend if there is further feedback. S3 has some knowledge gaps on the full feature scope, so we're meeting the known use case.

s3transfer/__init__.py
nateprewitt: I wrote a new test using the right infrastructure to make sure we've got this case covered.

s3transfer/upload.py
nateprewitt: Added it for good measure. After some testing it doesn't have any impact on the call so we might as well be complete and pass it if it was provided.

kyleknap (approved):
Looks good :ship:

PR #277, opened 2023-10-18
Add streaming fileobj support to CRTTransferManager

graebm:
Improvements to CRTTransferManager:
* Add streaming fileobj support
    * Previously, it only worked with file paths.
* Add support for `ChecksumAlgorithm`
* Use ALLOWED lists to limit what can be passed in `extra_args`
    * The default TransferManager uses these lists to prevent unexpected weirdness when splitting uploads and downloads into parts, but the CRTTransferManager hadn't hooked up these filters

### DIFF
--- a/s3transfer/crt.py
+++ b/s3transfer/crt.py
@@ -445,8 +445,17 @@ class BotocoreCRTRequestSerializer(BaseCRTRequestSerializer):
             # If host is not set, set it for the request before using CRT s3
             url_parts = urlsplit(botocore_http_request.url)
             crt_request.headers.set("host", url_parts.netloc)
+
         if crt_request.headers.get('Content-MD5') is not None:
             crt_request.headers.remove("Content-MD5")
+
+        # Explicitly set "Content-Length: 0" when there's no body.
+        # Botocore doesn't bother setting this, but CRT likes to know.
+        # Note that Content-Length SHOULD be absent if body is unseekable.
+        if crt_request.headers.get('Content-Length') is None:
+            if botocore_http_request.body is None:
+                crt_request.headers.add('Content-Length', "0")
+
         return crt_request
 
     def _capture_http_request(self, request, **kwargs):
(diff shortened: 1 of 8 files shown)


s3transfer/crt.py
```diff
@@ -150,6 +159,11 @@ def create_s3_crt_client(
 
 
 class CRTTransferManager:
+
+    ALLOWED_DOWNLOAD_ARGS = TransferManager.ALLOWED_DOWNLOAD_ARGS
+    ALLOWED_UPLOAD_ARGS = TransferManager.ALLOWED_UPLOAD_ARGS
+    ALLOWED_DELETE_ARGS = TransferManager.ALLOWED_DELETE_ARGS
```
graebm, line 165: Should I make a common base class for `TransferManager` and `CRTTransferManager`, where I can put stuff like these lists, and `def _validate_all_known_args()` and `def _validate_if_bucket_supported()`?

Or move these to be standalone lists/functions that both classes can use?

Or just copy/paste the functions, and reference the lists, like I'm doing here?

s3transfer/crt.py
```diff
@@ -556,22 +591,57 @@ def get_make_request_args(
...
+                send_filepath = call_args.fileobj
+                data_len = self._os_utils.get_file_size(send_filepath)
+                call_args.extra_args["ContentLength"] = data_len
+
+            elif call_args.fileobj is not None:
+                # fileobj is a file-like object
+                call_args.extra_args["Body"] = call_args.fileobj
+
+            # We want the CRT S3Client to calculate checksums, not botocore.
+            # Default to CRC32.
```
graebm, line 634: This is different than the default TransferManager. With this change CRT is defaulting to CRC32 instead of Content-MD5. The CRT team got grumpy about how Content-MD5 currently works in our code and pushed me to do it this way, but we can still make Content-MD5 work if we need to.

PR #279, opened 2023-11-02
Support file-like objects in CRT transfer manager

kyleknap:
This is a backport of this pull request: https://github.com/aws/aws-cli/pull/8291, which was implemented based off this pull request: https://github.com/boto/s3transfer/pull/277. Also note that the integration tests do not pass using the current version of CRT that is pinned in `botocore`. They do however pass if you update to the latest version of `awscrt`. I don't think that should block merging this PR as it will be merged into a feature branch. However, we will need to update the `awscrt` pin in `botocore` before merging the feature branch.

### DIFF
--- a/tests/integration/test_crt.py
+++ b/tests/integration/test_crt.py
@@ -11,11 +11,18 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import glob
+import io
 import os
 
 from s3transfer.subscribers import BaseSubscriber
 from s3transfer.utils import OSUtils
-from tests import HAS_CRT, assert_files_equal, requires_crt
+from tests import (
+    HAS_CRT,
+    NonSeekableReader,
+    NonSeekableWriter,
+    assert_files_equal,
+    requires_crt,
+)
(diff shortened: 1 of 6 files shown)


tests/integration/test_crt.py
```diff
@@ -44,20 +51,59 @@ def on_done(self, **kwargs):
...
+        self.download_path = os.path.join(self.files.rootdir, 'download.txt')
+
     def _create_s3_transfer(self):
         self.request_serializer = s3transfer.crt.BotocoreCRTRequestSerializer(
-            self.session
+            self.session, client_kwargs={'region_name': self.region}
         )
         credetial_resolver = self.session.get_component('credential_provider')
         self.s3_crt_client = s3transfer.crt.create_s3_crt_client(
-            self.session.get_config_variable("region"), credetial_resolver
```
kyleknap, line 53: It looks like this will pull in a fix that we made in v2 where we were using the session's region. However, we never directly set the region on the session even though we set the region to be `us-west-2` for the base test class.

nateprewitt (approved):
⛵

PR #281, opened 2023-11-14
Add CRT process lock utility

kyleknap:
The CRT S3 client performs best when there is only one instance of it running on a host. This lock allows an application to signal across processes whether there is another process of the same application using the CRT S3 client and prevent spawning more than one CRT S3 clients running on the system for that application.


nateprewitt (approved):
:shipit:

PR #284, opened 2023-11-24
Add error translation from S3ResponseError to Botocore errors

nateprewitt:
This PR adds functionality to the CRT S3 Transfer Manager to handle error translation from the S3 service. This will ensure Boto3 integration retains existing error hierarchy to provide a seamless experience between the two transfer methods.

Merge of this PR is pending https://github.com/boto/botocore/pull/3073 to provide the `operation_name` off the CRT shape.

### DIFF
--- a/s3transfer/crt.py
+++ b/s3transfer/crt.py
@@ -29,6 +29,7 @@ from awscrt.s3 import (
     S3Client,
     S3RequestTlsMode,
     S3RequestType,
+    S3ResponseError,
     get_recommended_throughput_target_gbps,
 )
 from botocore import UNSIGNED
@@ -196,6 +197,9 @@ class CRTTransferManager:
         self._s3_args_creator = S3ClientArgsCreator(
             crt_request_serializer, self._osutil
         )
+        self._crt_exception_translator = (
+            crt_request_serializer.translate_crt_exception
+        )
         self._future_coordinators = []
         self._semaphore = threading.Semaphore(128)  # not configurable
         # A counter to create unique id's for each transfer submitted.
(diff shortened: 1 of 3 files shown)


tests/unit/test_crt.py
```diff
@@ -164,6 +164,60 @@ def test_delete_request(self):
...
+                ('Content-Type', 'application/xml'),
+                ('Transfer-Encoding', 'chunked'),
+                ('Date', 'Fri, 10 Nov 2023 23:22:47 GMT'),
+                ('Server', 'AmazonS3'),
+            ],
+            body=body,
+            operation_name=operation_name,
+        )
+
+    def test_translate_get_object_404(self):
```
kyleknap, line 190: It might makes sense to port at least one of these test cases over to the functional test suite to ensure end to end that everything is working fine. Otherwise, we really only have the integration tests which are not run for GitHub CI.

s3transfer/crt.py
```diff
@@ -599,11 +646,27 @@ def result(self, timeout=None):
...
                 self._s3_request = None
-            self._crt_future.result(timeout)
+
+    def handle_exception(self, exc):
+        translated_exc = None
+        if self._exception_translator:
+            try:
+                translated_exc = self._exception_translator(exc)
+            except Exception:
+                # Bail out if we hit an issue translating
```
kyleknap, line 663: We should probably log if there was an exception and that we are ignoring it in favor the translated exception

kyleknap (commented):
Looks good. Just had a couple of comments.

tests/unit/test_crt.py
nateprewitt: Did a quick dive and this is possible but actually requires a bit more scaffolding that I expected. I'm going to vote we punt this part for now since we have an integ test that ensures this works end to end, and come back later to add a proper response-oriented functional test class.

kyleknap (approved):
:ship:

PR #291, opened 2023-11-29
Use lru_cache on costly validation function.

graebm:
This gave me a 24% speedup when downloading a folder with 30,000 small files using `aws s3 cp` with `preferred_transfer_client=crt` on a 100Gbps machine. It speeds up the `classic` transfer client as well.

When profiling this workload, `_validate_subscriber_methods()` had a big presence in the flame graph. Commenting out the validate function gave a huge speedup. But the validation is useful, so we didn't want to remove it. Using lru_cache provides the same speedup for the 30,000 file workload. It runs 6 times (once per subscriber class) instead of 180,000 times (6 times per file).


nateprewitt (approved):
:shipit:

PR #300, opened 2024-01-11
remove useless old six imports

a-detiste:
https://codesearch.debian.net/search?q=s3transfer.compat&literal=1&page=1&perpkg=1

### DIFF
--- a/s3transfer/__init__.py
+++ b/s3transfer/__init__.py
@@ -134,7 +134,6 @@ import socket
 import string
 import threading
 
-from botocore.compat import six  # noqa: F401
 from botocore.exceptions import IncompleteReadError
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hi @a-detiste, these are left in place for backwards compatibility. We don't intend to remove these at this time.

For the CLI v2 source, that is authored [here](https://github.com/aws/aws-cli/tree/v2/awscli/s3transfer). That's a more reasonable place these can be removed to help minimize the distro maintainer burden.

a-detiste:
This is an horrible mess, I'll look at it again later

a-detiste:
@nateprewitt awscli is now fixed. I think my PR still applies (and is still applied as a Debian downstream patch)

PR #301, opened 2024-03-06
Fix retrying incomplete downloads with urllib3 2.x

apmorton:
https://github.com/boto/botocore/issues/3132


nateprewitt (approved):
Thanks for the PR, @apmorton! Yep, this definitely looks like a missed behavior change when 2.x support was added. We'd be happy to merge this. I made a minor change to fix a linting error, but otherwise this looks good.

PR #309, opened 2024-06-21
Pass operation name to awscrt.s3

graebm:
`awscrt.s3` will soon require that `operation_name` be passed for `S3RequestType.DEFAULT` (see: https://github.com/awslabs/aws-c-s3/pull/439, https://github.com/awslabs/aws-crt-python/pull/574)

Only `CRTTransferManager.delete()` uses `S3RequestType.DEFAULT`. Fix it to pass along `operation_name="DeleteObject"`

### DIFF
--- a/.changes/next-release/bugfix-awscrt-54174.json
+++ b/.changes/next-release/bugfix-awscrt-54174.json
@@ -1,5 +1,5 @@
 {
   "type": "bugfix",
-  "category": "awscrt",
-  "description": "Pass operation name to awscrt.s3, to improve error handling"
+  "category": "``awscrt``",
+  "description": "Pass operation name to awscrt.s3 to improve error handling."
 }
--- a/.changes/next-release/bugfix-awscrt-54174.json
+++ b/.changes/next-release/bugfix-awscrt-54174.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "awscrt",
+  "description": "Pass operation name to awscrt.s3, to improve error handling"
+}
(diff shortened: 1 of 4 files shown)


s3transfer/crt.py
```diff
@@ -818,13 +818,22 @@ def _default_get_make_request_args(
...
+
         make_request_args = {
             'request': self._request_serializer.serialize_http_request(
                 request_type, future
             ),
-            'type': getattr(
-                S3RequestType, request_type.upper(), S3RequestType.DEFAULT
-            ),
+            'type': crt_request_type,
+            'operation_name': operation_name,
```
nateprewitt, line 836: How is this going to work for older releases on platforms like the P4D and TRN1 instances? Do the existing releases already allow this argument but it's not required?

graebm: yeah, the argument has been there since last year, but it was optional and we forget to use it here.

graebm: This was one of the last things we added last year for the boto3 integration was: https://github.com/awslabs/aws-crt-python/pull/530 Add S3ResponseError.operation_name

> Also, if user is creating a DEFAULT request (where CRT doesn't know which operation is being performed), the user can pass in the operation_name

.changes/next-release/bugfix-awscrt-54174.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "awscrt",
+  "description": "Pass operation name to awscrt.s3, to improve error handling"
```
nateprewitt, line 4:
```suggestion
  "category": "``awscrt``",
  "description": "Pass operation name to awscrt.s3 to improve error handling."
```

nateprewitt (approved):

PR #316, opened 2024-10-17
Skip HEAD for multipart check when multipart_threshold is 0

daveisfera:
Each download does a `HEAD` before the `GET`, but unless the file is possible a very large file that would benefit from multipart download, then this is just adding unneeded operations and delay. This PR changes it so that `multipart_threshold` being 0 will skip the `HEAD` call and I'd argue that this should be the changed to the default behavior as well

Related to https://github.com/boto/boto3/issues/3083

### DIFF
--- a/s3transfer/__init__.py
+++ b/s3transfer/__init__.py
@@ -747,6 +747,7 @@ class S3Transfer:
             unique_id='s3upload-callback-enable',
         )
         if (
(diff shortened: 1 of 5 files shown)


daveisfera:
Changed to `None` since there's a check that the value isn't `0`

PR #325, opened 2025-01-16
Update floor for botocore

nateprewitt:
This fixes conflicts with imports when using older versions of botocore. With the recent changes in 0.11.0, botocore 1.36.0+ is now required.


jonathan343 (approved):
:shipit:

Issue #327, opened 2025-01-17
`aws s3 cp` does not honor `request_checksum_calculation = WHEN_REQUIRED`


metadaddy:
When using the AWS CLI with a third-party S3-compatible service, such as [Backblaze B2](https://www.backblaze.com/cloud-storage), `aws s3 cp` sends the `x-amz-sdk-checksum-algorithm` and `x-amz-checksum-crc64nvme` headers, even if `request_checksum_calculation` is set to `WHEN_REQUIRED` in the configuration.

[Data Integrity Protections for Amazon S3](https://docs.aws.amazon.com/sdkref/latest/guide/feature-dataintegrity.html) states that, when `request_checksum_calculation` is set to `WHEN_REQUIRED`, 'Checksum validation is performed only when required by the API operation.'

`.aws/config`

```
[profile b2]
endpoint_url = https://s3.us-west-004.backblazeb2.com
region = us-west-004
request_checksum_calculation = WHEN_REQUIRED
response_checksum_validation = WHEN_REQUIRED
```

Running the CLI:

```
% aws --profile b2 s3 cp ./hello.txt s3://metadaddy-private/
upload failed: ./hello.txt to s3://metadaddy-private/hello.txt An error occurred (InvalidArgument) when calling the PutObject operation: Unsupported header 'x-amz-sdk-checksum-algorithm' received for this API call.
```

This is happening because S3 Transfer Manager ensures that `ChecksumAlgorithm` is set, either to the user-specified value, or `CRC64NVME` (see [the `set_default_checksum_algorithm()` function](https://github.com/boto/s3transfer/blob/cb67175a34448e5d10458af4857f8282f313c167/s3transfer/utils.py#L829), added just a couple of days ago).

Now, in [Botocore's `resolve_request_checksum_algorithm()` function](https://github.com/boto/botocore/blob/develop/botocore/httpchecksum.py#L256), `request_checksum_required` evaluates to `False`, as you might expect, but, since `ChecksumAlgorithm` is set in the incoming params, `if algorithm_member and algorithm_member in params:` evaluates to `True` and the checksum headers are calculated and sent in the request.

I think `TransferManager._add_operation_defaults()` should only call `set_default_checksum_algorithm()` if `request_checksum_calculation` is set to `WHEN_SUPPORTED`. Something like this:

```python
    def _add_operation_defaults(self, extra_args):
        if self.client.meta.config.request_checksum_calculation == "when_supported":
[...]

PR #328, opened 2025-01-18
Honor "request_checksum_calculation = WHEN_REQUIRED" in config (#327)

metadaddy:
As mentioned in #327, `aws s3 cp` does not honor the `request_checksum_calculation` configuration, sending an `x-amz-sdk-checksum-algorithm` header even when `request_checksum_calculation` is set to `WHEN_REQUIRED`.

This change adds a check to `TransferManager._add_operation_defaults()` so that it only calls `set_default_checksum_algorithm()` if `request_checksum_calculation` is set to `WHEN_SUPPORTED`.

### DIFF
--- a/s3transfer/manager.py
+++ b/s3transfer/manager.py
@@ -519,7 +519,8 @@ class TransferManager:
                 )
 
     def _add_operation_defaults(self, extra_args):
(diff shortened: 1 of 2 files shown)


metadaddy:
Closing, since this was implemented as #329

PR #335, opened 2025-02-13
User agent 2.1: Propagate Botocore context to child threads

hssyoo:
Companion PR to https://github.com/boto/botocore/pull/3389

#### Description
Context variables aren't automatically propagated from parent to child threads. To ensure the Botocore client captures features registered outside of concurrent context, we need to explicitly get and pass the context to the thread pool executor.

### DIFF
--- a/s3transfer/futures.py
+++ b/s3transfer/futures.py
@@ -17,6 +17,8 @@ import threading
 from collections import namedtuple
 from concurrent import futures
 
+from botocore.context import get_context
+
 from s3transfer.compat import MAXINT
 from s3transfer.exceptions import CancelledError, TransferNotDoneError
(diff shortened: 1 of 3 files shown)


s3transfer/futures.py
```diff
@@ -17,6 +17,8 @@
 from collections import namedtuple
 from concurrent import futures
 
+from botocore.context import get_context
```
jonathan343, line 20: The current version requirement for s3transfer is:

```python
requires = [
    'botocore>=1.36.0,<2.0a.0',
]
```

We should bump the minimum supported version to prevent import errors since the `botocore.context` won't exist in versions between 1.36.0 and the version that releases https://github.com/boto/botocore/pull/3389

jonathan343 (approved):

PR #358, opened 2025-08-22
Validate multipart download content ranges

hssyoo:
Reverse port of https://github.com/aws/aws-cli/pull/9666

Add runtime validation to ensure that the fetched content ranges match the requested ranges for multipart downloads.


jonathan343 (commented):
Changes LGTM! I'm thinking this warrants a changelog entry though. I also think it's a good idea to ship this in a new minor version

jonathan343 (approved):
:shipit:

PR #363, opened 2025-11-21
Eliminate HEAD requests during downloads, especially for faster transfers of small files

crowecawcaw:
This PR optimizes downloads through the `TransferManager` by removing the upfront HEAD request on an opt-in basis. Previously, every download issued a HEAD request to determine object size before starting the GET. This change eliminates that extra round-trip by extracting metadata from the first GET response instead, **when the caller has configured `response_checksum_validation='when_required'` on the client**. The default path still uses HEAD so that full-object checksum validation continues to work.

For small files, the download time is dominated by request latency, so eliminating one of the two requests results in a ~50% download time reduction. For large files, the effect is less noticeable because the download time is dominated by the transfer itself and there are multiple chunks. In both cases, we save the cost of the HEAD request.

### What Changed

• **No-HEAD download path added, gated on `response_checksum_validation='when_required'`**: Downloads on that path start immediately with a ranged GET request for the first chunk. The default path is unchanged.
• **Dynamic size detection (no-HEAD path)**: Extract object size and ETag from the first GET response headers (`ContentRange` or `ContentLength`).
• **Dynamic chunk scheduling (no-HEAD path)**: After the first chunk completes, schedule additional chunks only if the object is larger than the chunk size.
• **ETag consistency for all chunks**: When an ETag is available (either pre-provided via a subscriber, e.g. from a prior HEAD request by the AWS CLI, or extracted from the first GET response), all subsequent ranged GET requests include an `IfMatch` header with that ETag. This ensures S3 rejects the request if the object changes mid-download. The first GET request also includes `IfMatch` if an ETag was pre-provided before the download started.
• **Zero-byte object handling (no-HEAD path)**: The first ranged GET on a 0-byte object returns `InvalidRange`. The new path handles this by returning an empty response instead of retrying with a non-ranged GET.

### Testing

Unit, functional, and integ tests pass. To produce the timing numbers below, I used a small benchmarking script that downloads many small files in a batch — it isn't included in the PR since `scripts/` contains inactive utilities, but it is viewable on my fork at [`scripts/performance/time-batch-download.py` @ c482815](https://github.com/crowecawcaw/s3transfer/blob/c482815/scripts/performance/time-batch-download.py). For 1000 × 1 KB files on my laptop, total download duration dropped ~41% (from 15.0s to 8.9s) on the opt-in path.

### Backward Compatibility

External API unchanged. All download methods have the same signatures. The default path (HEAD + size-branching GET) is preserved, so no behavior change for callers who don't set `response_checksum_validation='when_required'`.

#### Default path (unchanged)

```mermaid
flowchart TD
    A[HEAD Request] --> C{Size < 8MB?}
    C -->|Yes| D[GET Request]
    C -->|No| E[Multiple GET Requests]
    D --> F[Complete]
    E --> F
```

#### No-HEAD path (opt-in via response_checksum_validation="when_required")

```mermaid
flowchart TD
    A[GET First Chunk] --> B{Size < 8MB?}
    B -->|Yes| C[Complete]
    B -->|No| D[GET Remaining Chunks]
    D --> C
```


s3transfer/download.py
```diff
@@ -424,123 +392,182 @@ def _submit_download_request(
...
-                },
-                done_callbacks=[final_task],
-            ),
+            task,
             tag=get_object_tag,
         )
 
-    def _submit_ranged_download_request(
+
+class GetObjectOnDoneCallback:
```
aemous: nit: The callback is run after the first `GetObject` call and schedules remaining work based on the response. Can we use a more specific name?

maybe `GetObjectFirstChunkOnDoneCallback`?

aemous (commented):
I left 1 nit. Note that I am not the primary reviewer for this PR, and we are still pending the primary review.

s3transfer/download.py
```diff
@@ -424,123 +392,182 @@ def _submit_download_request(
...
-
-            # Inject extra parameters to be passed in as extra args
             extra_args = {
                 'Range': range_parameter,
             }
-            if transfer_future.meta.etag is not None:
-                extra_args['IfMatch'] = transfer_future.meta.etag
+            # Use IfMatch to ensure object hasn't changed during download
+            if etag is not None:
+                extra_args['IfMatch'] = etag
```
aemous, line 553: So each ranged GetObject except the first will specify an etag in the request headers. Can we add this as a change to the PR description?

s3transfer/download.py
```diff
@@ -424,123 +392,182 @@ def _submit_download_request(
         # Get any associated tags for the get object task.
         get_object_tag = download_output_manager.get_download_task_tag()
 
-        # Get the final io task to run once the download is complete.
-        final_task = download_output_manager.get_final_io_task()
+        # Request first chunk to get object metadata from response headers
+        chunk_size = config.multipart_chunksize
+        extra_args = dict(call_args.extra_args)
+        extra_args['Range'] = f'bytes=0-{chunk_size - 1}'
```
aemous, line 398: The AWS CLI performs a HEAD request before submitting the GetObject request to s3transfer, and stores the `etag` and size from the HEAD request on the `transfer_future`.

Can we put the `etag` on the first GetObject request if one is provided so that if the object changes between the HEAD request performed by the AWS CLI and the first GetObject by s3transfer, the operation will be rejected by S3?

```python
if transfer_future.meta.etag is not None:
    extra_args['IfMatch'] = transfer_future.meta.etag
```

aemous:
Looks like downloads may be failing when the object has a size of 0 bytes. The error I'm seeing is 

```
botocore.exceptions.ClientError: An error occurred (InvalidRange) when calling the GetObject operation: The requested range is not satisfiable
```

Can you confirm whether you are able to reproduce this, and look into resolving it?

InvalidRange occurs when the range we are requesting has no overlap with the object itself. For 0 byte objects, this will always be the case if we are requesting the first 0-x bytes, since there are no bytes at all. The only feasible patch I can imagine here would be a try-except clause that handles InvalidRange by performing a non-ranged GET, which may be undesirable. If you have a better alternative for resolving this that doesn't involve needing an extra request, that would be preferred.

crowecawcaw:
I see the same issue. I added an integ test, fixed the unit test to mirror the same behavior, then fixed the issue. Instead of sending a second Get request for an object we know is empty, it just returns the empty content.

aemous:
@crowecawcaw Can you push a revision that satisfies the 'Lint code' GH Action?

In case you cannot see the Action details, I pasted it here:

```
pre-commit hook(s) made changes.
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.
All changes made by hooks:
diff --git a/scripts/performance/time-batch-download.py b/scripts/performance/time-batch-download.py
index 93e8a2a..1cf04fc 100755
--- a/scripts/performance/time-batch-download.py
+++ b/scripts/performance/time-batch-download.py
@@ -2,10 +2,12 @@
 """Direct timing of batch downloads without shell wrapper."""
 
 import argparse
-import tempfile
 import shutil
+import tempfile
 import time
+
 from botocore.session import get_session
+
 
 
@@ -20,13 +22,13 @@ def main():
     parser.add_argument('--file-size', type=int, required=True)
     parser.add_argument('--s3-bucket', required=True)
     args = parser.parse_args()
-    
+
     session = get_session()
     client = session.create_client('s3')
-    
+
     tempdir = tempfile.mkdtemp()
     s3_keys = []
-    
+
     try:
         # Upload files
         print(f"Uploading {args.file_count} files...")
@@ -37,7 +39,7 @@ def main():
                 s3_key = f"perf_test_{i}"
                 manager.upload(file_path, args.s3_bucket, s3_key)
                 s3_keys.append(s3_key)
-        
+
         # Download files
         print(f"Downloading {args.file_count} files...")
         start_time = time.time()
@@ -46,9 +48,9 @@ def main():
                 download_path = f"{tempdir}/download_{i}"
                 manager.download(args.s3_bucket, s3_key, download_path)
         duration = time.time() - start_time
-        
+
         print(f"Download duration: {duration:.2f} seconds")
-        
+
         # Cleanup
         for s3_key in s3_keys:
             client.delete_object(Bucket=args.s3_bucket, Key=s3_key)
```

.changes/next-release/feature-download-85557.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
```
aemous: Can we regenerate the changelog entry to use the `enhancement` type? Feature results in a minor version bump, and we try to restrict it to extremely large changes.

Also, I recommend ``s3`` as a category instead.

And can we update the description to mention S3 up-front? e.g. Skip HEAD requests during S3 downloads when ...

s3transfer/download.py
```diff
@@ -541,6 +557,216 @@ def _calculate_range_param(self, part_size, part_index, num_parts):
...
+        self._request_executor = request_executor
+        self._bandwidth_limiter = bandwidth_limiter
+        self._fileobj = fileobj
+        self._progress_callbacks = progress_callbacks
+        self._get_object_tag = get_object_tag
+        self._task = None
+
+    def __call__(self):
+        # Always check if we have a task and response first
+        if not self._task:
```
aemous: What cases do we expect the task to be None? Or is this check just for defensive programming purposes?

scripts/performance/time-batch-download.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
```
aemous: Has this script been used to produce benchmarking results that summarize the latency improvement this PR gives?

tests/unit/test_download.py
```diff
@@ -1044,3 +1045,77 @@ def test_duplicate_writes_longer_length_update_queue(self):
...
+        callback()
+        self._assert_only_final_task_submitted()
+
+
+class TestValidateContentRange(unittest.TestCase):
+    """Covers edge-case early returns in GetObjectTask._validate_content_range."""
+
+    def _validator(self):
+        # Bind the unbound method; validator doesn't touch instance state.
+        return GetObjectTask._validate_content_range.__get__(
```
aemous: Is there any good way to test this functionality without accessing private functions/variables of another class? For example, maybe using the public interface that indirectly calls these private functions to indirectly test this underlying behavior?

aemous (changes requested):
Loogs good overall. Left a few questions/comments.

s3transfer/download.py
crowecawcaw: Just defensive, I'll simplify this to an assert.

tests/unit/test_download.py
crowecawcaw: Updated!

scripts/performance/time-batch-download.py
crowecawcaw: Yes, it shows ~40% reduction in transfer time for small files with checksumming not required (details below). I imagined it could be added to a automated benchmark suite later. For now, I just added it as a manually run script to show how to measure the improvement.


```
Benchmark Results — s3transfer PR #363

  Test: Download 1000 × 1KB files from EC2 instance to S3 bucket in us-east-1

  ┌──────────────────────┬────────────────────────────────────────────┬───────────────┬───────────────────┐
  │ Mode                 │ Config                                     │ Download Time │ Requests per file │
  ├──────────────────────┼────────────────────────────────────────────┼───────────────┼───────────────────┤
  │ Default (HEAD + GET) │ response_checksum_validation unset         │ 6.11s         │ 2 (HEAD then GET) │
  ├──────────────────────┼────────────────────────────────────────────┼───────────────┼───────────────────┤
  │ Optimized (GET only) │ response_checksum_validation=when_required │ 3.51s         │ 1 (GET only)      │
  └──────────────────────┴────────────────────────────────────────────┴───────────────┴───────────────────┘

  Improvement: 42.6% faster (2.60s saved) when skipping the HEAD request.
```

s3transfer/download.py
```diff
@@ -541,6 +557,214 @@ def _calculate_range_param(self, part_size, part_index, num_parts):
...
+        self._bandwidth_limiter = bandwidth_limiter
+        self._fileobj = fileobj
+        self._progress_callbacks = progress_callbacks
+        self._get_object_tag = get_object_tag
+        self._task = None
+
+    def __call__(self):
+        # Always check if we have a task and response first
+        assert self._task is not None, (
+            "set_task() must be called before the task is submitted"
```
aemous: Can we replace this assertion with raising a `RuntimeError` with a similar message wrapped? It would be more idiomatic with our codebase.

aemous (changes requested):
Looks good overall, 1 small change requested.

scripts/performance/time-batch-download.py
aemous: Nice!

I agree this can be useful as an automated benchmark suite in the future.

s3transfer/download.py
```diff
@@ -541,6 +557,214 @@ def _calculate_range_param(self, part_size, part_index, num_parts):
...
+        self._task = None
+
+    def __call__(self):
+        if self._task is None:
+            raise RuntimeError(
+                "set_task() must be called before the task is submitted"
+            )
+
+        response = self._task.get_response()
+        if not response:
```
aemous: nit: this is likely fine, but it may be worth being more explicit in what we are checking against here. 

Is this if-statement checking for a None response?
Is it checking for an empty body response?
Is it checking for any arbitrary falsy value?

Currently it is checking for any falsy value. I'm okay with this, but if we know for sure we are checking for one specific value, then I would prefer checking for that value expicitly.

scripts/performance/time-batch-download.py
```diff
@@ -0,0 +1,62 @@
...
+                file_path = f"{tempdir}/upload_{i}"
+                create_file(file_path, args.file_size)
+                s3_key = f"perf_test_{i}"
+                manager.upload(file_path, args.s3_bucket, s3_key)
+                s3_keys.append(s3_key)
+
+        # Download files
+        print(f"Downloading {args.file_count} files...")
+        start_time = time.time()
+        with TransferManager(client) as manager:
```
aemous: nit: do we need to initialize a new TransferManager here or can we reuse the previous one? Especially because it is after the timer started.

tests/integration/test_download.py
```diff
@@ -282,3 +303,26 @@ def test_download_to_special_file(self):
                 'Should have been able to download to /dev/null but received '
                 f'following exception {e}'
             )
+
+
+class TestDownloadWhenRequired(TestDownload):
```
aemous: nit: Without context, this class name is not descriptive. I may prefer something like

```python
TestDownloadResponseChecksumValidationWhenRequired
```

tests/unit/test_download.py
```diff
@@ -1044,3 +1045,55 @@ def test_duplicate_writes_longer_length_update_queue(self):
                 {'offset': 1, 'data': 'bar'},
             ],
         )
+
+
+class TestGetObjectFirstChunkOnDoneCallback(unittest.TestCase):
```
aemous: For net-new test suites, instead of extending `unittest.TestCase`, lets instead use PyTest.

aemous (changes requested):
Requested 1 small change and 3 non-blocking nits. Leaving the nits up to your discretion.

aemous (changes requested):
Code looks good and manual testing has succeeded.

Can you please delete the new benchmarking script? We do not use those benchmarking scripts under `scripts/performance` currently, and would rather not add a new script that we don't use. 

You can include your script as an attachment in the GH PR description, and we may reference it in the future.

aemous (approved):
LGTM. 

I performed some manual testing where I downloaded a file from S3 with all possible checksum-validation options, and I also tried with/without overriding `ChecksumMode=ENABLED` at the `GetObject` operation level.

PR #372, opened 2026-03-12
Raise TypeError on bare test decorator usage, fix bare @requires_crt usages, and add regression tests

Alan4506:
*Background:*
`@skip_if_windows` and `@skip_if_using_serial_implementation` are decorator factories that require parentheses. If they are used without parentheses (e.g., bare `@skip_if_windows`), this is unsafe because Python passes the test function as the first argument, causing the real test body to never run. Assertions are never evaluated, but the test may still appear to pass or skip silently. However, these two did not raise an error on bare usage.

On the other hand, `@requires_crt` in s3transfer only worked correctly with bare usage, while botocore and boto3 already required the parenthesized form `@requires_crt()`. This cross-repo inconsistency creates a maintainer footgun because contributors move between repos and assume the decorator behaves the same way everywhere. It needs to be standardized to only accept the parenthesized form and all existing bare usages need to be updated.

*Description of changes:*
`@skip_if_windows` and `@skip_if_using_serial_implementation` now raise a `TypeError` on bare usage. `@requires_crt` is updated to only accept the parenthesized form for consistency with the other two decorators in this repo as well as botocore and boto3. It also raises a `TypeError` on bare usage. Five existing bare `@requires_crt` usages across test files are updated to `@requires_crt()`.

*Testing:*
Regression tests are added in `tests/unit/test_decorators.py` to verify correct behavior for all three decorators. All tests including the new tests pass.

### DIFF
--- a/tests/unit/test_decorators.py
+++ b/tests/unit/test_decorators.py
@@ -1,0 +1,108 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the 'License'). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the 'license' file accompanying this file. This file is
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from unittest import mock
+
+import pytest
+
+from tests import (
+    requires_crt,
+    skip_if_using_serial_implementation,
+    skip_if_windows,
+)
+
+
+class TestSkipIfWindows:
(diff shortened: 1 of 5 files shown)


jonathan343 (changes requested):
Looks good to me but the linter is failing. If you install pre-commit, you can automatically have your changes linted and formatted before they're committed. You can do this by running `uv tool install pre-commit`. Then `pre-commit install`. For a one time fix, you can manually run the checks with `pre-commit run --all-files`.

Note: `prek` is a new drop-in replacement for pre-commit that I've been using. If you use this, replace `pre-commit` in the commands I posted above with `prek`

tests/unit/test_decorators.py
```diff
@@ -0,0 +1,108 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
alexgromero: We should remove the year here

jonathan343 (approved):

PR #389, opened 2026-05-04
Use SPDX license identifier

Shortfinga:
The SPDX ID for the Apache 2 is Apache-2.0. This would help compliance and license scanners to determine the correct license.


setup.py
btschwertfeger: This could be removed as well, see [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-classifiers)

btschwertfeger:
PUSH: I would also like to see this change merged and released. :+1:

setup.py
Shortfinga: Done

PR #391, opened 2026-05-08
Add Free Threading support to CI

nateprewitt:
*Description of changes:*
The primary blocker for supporting [free threading](https://peps.python.org/pep-0779/) in Boto3/Botocore/Smithy Python was support in the upstream dependency, awscrt. Starting in awscrt 0.32.0 (https://github.com/awslabs/aws-crt-python/pull/725), releases began distributing artifacts for Python 3.13t and 3.14t. While 3.13t was released as experimental, 3.14t is starting to gain adoption.

Given there are no remaining blockers, this PR adds beta support for free threaded builds and makes sure that contract is validated in CI.

### DIFF
--- a/.github/workflows/run-crt-test.yml
+++ b/.github/workflows/run-crt-test.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


TingDaoK (commented):
shipit!!!!!

jonathan343 (approved):

PR #392, opened 2026-05-26
Preserve source object metadata for multipart copies

hssyoo:
Extract metadata fields (that get copied automatically for single `CopyObject` requests) from initial `HeadObject` call and pass them into multipart copies.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html


SamRemis (approved):
:shipit:

PR #410, opened 2026-07-18
Raise error on empty CA bundle value

alexgromero:
*Issue #, if available:*

*Description of changes:*
When a CA bundle value (`ca_bundle`, `AWS_CA_BUNDLE`, `REQUESTS_CA_BUNDLE`, or the `verify` parameter) resolves to an empty string, `create_s3_crt_client` will now raise an error.

An empty string is not a valid value for these settings, which are defined as paths to a CA bundle.

### DIFF
--- a/.changes/next-release/bugfix-TLS-36922.json
+++ b/.changes/next-release/bugfix-TLS-36922.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "TLS",
+  "description": "Return a configuration error when the CA bundle value (ca_bundle, AWS_CA_BUNDLE, REQUESTS_CA_BUNDLE, or verify) resolves to an empty string."
+}
--- a/.changes/next-release/bugfix-TLS-36922.json
+++ b/.changes/next-release/bugfix-TLS-36922.json
@@ -1,5 +1,5 @@
 {
   "type": "bugfix",
   "category": "TLS",
-  "description": "Return a configuration error when the CA bundle value (ca_bundle, AWS_CA_BUNDLE, REQUESTS_CA_BUNDLE, or verify) resolves to an empty string."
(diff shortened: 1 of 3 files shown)


ubaskota (approved):
LGTM!

ubaskota (commented):
Quick Note: `.strip()` is only used for the emptiness check, not to normalize the value. Before this change, a whitespace-only string like `'   '` reached `open()` [in `_read_binary_file`](https://github.com/awslabs/aws-crt-python/blob/5e7eef180f56818a3777d644f97b5478f02bd7c9/awscrt/io.py#L210) and raised a `FileNotFoundError`, now it raises `InvalidConfigError`, upfront.

I'll approve it as it is but one quick question: What are your thoughts on stripping `verify` in all cases vs. just the emptiness check? For example: what will happen with `verify="path/to/sth.sth   "`?

alexgromero:
> Quick Note: `.strip()` is only used for the emptiness check, not to normalize the value. Before this change, a whitespace-only string like `' '` reached `open()` [in `_read_binary_file`](https://github.com/awslabs/aws-crt-python/blob/5e7eef180f56818a3777d644f97b5478f02bd7c9/awscrt/io.py#L210) and raised a `FileNotFoundError`, now it raises `InvalidConfigError`, upfront.
> 
> I'll approve it as it is but one quick question: What are your thoughts on stripping `verify` in all cases vs. just the emptiness check? For example: what will happen with `verify="path/to/sth.sth "`?

I'd keep it to the emptiness check and not normalize. Leading/trailing spaces are technically valid in filenames, so stripping could make a legitimately named bundle inaccessible. Better to pass the path through as-is and let open() decide. "path/to/sth.sth   " would raise `FileNotFoundError`, which is the right fail-closed behavior.

